PR #2, opened 2022-04-19
Initial codebase review for FireFly Transaction Manager

peterbroadhurst:
This is a unit tested base set of functionality for the FireFly transaction manager for review.

Aligns with high level design outlined in https://github.com/hyperledger/firefly-ethconnect/issues/149#issuecomment-920866244

The architecture is summarized as follows:

- New **FireFly Transaction Manager** runtime
  - Operates as a microservice extension of the FireFly Core
  - Uses the `operation` resource within FireFly Core to store and update state
  - Runs as a singleton and is responsible for `nonce` assignment
  - Takes as much heavy lifting away from blockchain specific connectors as possible
- Lightweight FireFly Connector API (`ffcapi`)
  - Simple synchronous RPC operations that map to the most common operations supported across public blockchain technologies
  - Examples:
    - Find the next nonce for a given signing key
    - Serialize a transaction from a set of JSON inputs and an interface definition
    - Submit an un-signed transaction with a given gas price to the blockchain, via a signing wallet
    - Establish a new block listener
    - Poll for new blocks
    - Establish a new event log listener
    - Poll for new events
- Pluggable **Policy Engine**
  - Invoked to make decisions on transaction submission
  - Responsible for gas price calculation
  - Able to intervene and adjust the characteristics of signing/submission
  - OSS reference implementation provided with Gas Station REST API integration
- **Confirmation Manager**
  - Extracted from the Ethconnect codebase
  - Coupled to both transaction submission and event confirmation
  - Embeds an efficient block cache
- **Event Streams** - _not yet completed
  - Extracted from the Ethconnect codebase
  - Checkpoint restart based reliable at-least-once delivery of events
  - WebSockets interface upstream to FireFly Core


awrichar:
Thanks for posting this - I've gone over the high level and it seems to match what we've discussed for this piece. I'm still going over the changes in more detail to see if I have any questions/concerns, but so far looks like a very good baseline.

PR #5, opened 2022-04-26
Block based receipt detection in confirmation mgr, and optional change listener

peterbroadhurst:
In PR chain with #4 

- Adds `transactionHashes` to block info on `ffcapi`
- Moves receipt detection to confirmation manager
  - Checks once explicitly when TX added to confirmation manager
  - Then by detecting inclusion in block hashes
- Makes the change listener optional
  - Not needed when FFTM is the only component updating these operations in FF Core (common case)
- Updates the config for clarity for above


internal/policyengines/simple/simple_policy_engine.go
```diff
@@ -46,20 +46,24 @@ func (f *PolicyEngineFactory) Name() string {
...
+		gasOracleQueryInterval: gasOraclePrefix.GetDuration(GasOracleQueryInterval),
+		gasOracleMode:          gasOraclePrefix.GetString(GasOracleMode),
 	}
-	if gasStationEnabled {
-		p.gasStationClient = ffresty.New(ctx, gasStationPrefix)
-	}
-	if p.fixedGasPrice.IsNil() && p.gasStationClient == nil {
-		return nil, i18n.NewError(ctx, tmmsgs.MsgNoGasConfigSetForPolicyEngine, prefix.Resolve(FixedGasPrice), gasStationPrefix.Resolve(GasStationEnabled))
+	switch p.gasOracleMode {
+	case GasOracleModeConnector:
```
awrichar, line 60: Is this branch intentionally empty?

awrichar (approved):
Looks really good - left a couple of questions but I don't think there are any issues.

internal/manager/changelistener.go
```diff
@@ -0,0 +1,87 @@
...
+	log.L(m.ctx).Debugf("%s:%s/%s operation change event received", ce.Namespace, ce.ID, ce.Type)
+	if ce.Collection == "operations" && ce.Type == fftypes.ChangeEventTypeUpdated {
+		m.mux.Lock()
+		_, knownID := m.pendingOpsByID[*ce.ID]
+		m.mux.Unlock()
+		if !knownID {
+			// Currently the only action taken for change events, is to check we are
+			// tracking the transaction. However, as only transactions we submitted are
+			// valid and we do a full query on startup - the change listener is a little
+			// redundant (and disabled by default)
```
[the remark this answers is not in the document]
peterbroadhurst: Merits further discussion. I feel like it will, as I think there's going to be cases where we need to change operation outside of the FFTM runtime. However, that has its own concerns as the policy engine doesn't have any way to do optimistic concurrency locking on it' changes to the operations today. So I'm honestly not sure.

internal/policyengines/simple/simple_policy_engine.go
peterbroadhurst: added a comment

PR #7, opened 2022-06-02
Refactor FFCAPI to a local API interace, and FFTM to a utility package

peterbroadhurst:
This is quite a big change.

- Removes the Docker build from this package - this means the existing builds will stay for those using them
- Reinstates the FFCAPI definition here, which was recently moved to FF common, and now removed in https://github.com/hyperledger/firefly-common/pull/15
- Refactors the FFCAPI to be a local interface, rather than remote
   - This is because it's just too complex to having the potential of state mismatch in event state between the FFCAPI and the FFTM, and build a clean interface
 - Removes the `fftm` and `cmd` directories
 - Moves the `internal/manager` to be `pkg/fftm` for use by FFCAPI connectors
   - They will inject their implementation into the `fftm.Manager` to start their server


peterbroadhurst:
Moving to draft until we have confirmed correct operation with the latest `alpha` code and the old FFTM codestream

pkg/fftm/route__root_command.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+var postRootCommand = func(m *manager) *ffapi.Route {
+	return &ffapi.Route{
+		Name:            "postRootCommand",
+		Path:            "/",
+		Method:          http.MethodPost,
+		PathParams:      nil,
+		QueryParams:     nil,
+		Description:     tmmsgs.APIEndpointPostSubscriptions,
```
nguyer, line 38: This doesn't look like the correct description for this route

pkg/fftm/route__root_command.go
```diff
@@ -0,0 +1,81 @@
...
+			if err == nil {
+				schemas = append(schemas, queryRequest)
+			}
+			return &openapi3.SchemaRef{
+				Value: &openapi3.Schema{
+					AnyOf: schemas,
+				},
+			}, err
+		},
+		JSONOutputCodes: []int{http.StatusOK},
```
nguyer, line 57: For async requests, should this return `202 Accepted`, instead of `200 OK`? Not sure if that will cause a compatibility issue with existing projects.

pkg/fftm/route_post_subscription_reset.go
```diff
@@ -0,0 +1,45 @@
...
+	return &ffapi.Route{
+		Name:       "postSubscriptionReset",
+		Path:       "/subscriptions/{listenerId}/reset",
+		Deprecated: true, // in favor of "/eventstreams/{streamId}/listeners/{listenerId}"
+		Method:     http.MethodPost,
+		PathParams: []*ffapi.PathParam{
+			{Name: "listenerId", Description: tmmsgs.APIParamListenerID},
+		},
+		QueryParams:     nil,
+		Description:     tmmsgs.APIEndpointPatchSubscription,
```
nguyer, line 37: Looks like this (and maybe the comment above) are pointing to a description for a different endpoint?

pkg/fftm/route_post_subscriptions.go
```diff
@@ -0,0 +1,46 @@
...
+var postSubscriptions = func(m *manager) *ffapi.Route {
+	return &ffapi.Route{
+		Name:       "postSubscriptions",
+		Path:       "/subscriptions",
+		Deprecated: true, // in favor of "/eventstreams/{id}/listeners"
+		Method:     http.MethodPost,
+		PathParams: nil,
+		QueryParams: []*ffapi.QueryParam{
+			{Name: "limit", Description: tmmsgs.APIParamLimit},
+			{Name: "after", Description: tmmsgs.APIParamAfter},
```
nguyer, line 36: Are these applicable for creating a listener?

pkg/fftm/route_post_eventstream_listeners.go
```diff
@@ -0,0 +1,47 @@
...
+	return &ffapi.Route{
+		Name:   "postEventStreamListeners",
+		Path:   "/eventstreams/{streamId}/listeners",
+		Method: http.MethodPost,
+		PathParams: []*ffapi.PathParam{
+			{Name: "streamId", Description: tmmsgs.APIParamStreamID},
+		},
+		QueryParams: []*ffapi.QueryParam{
+			{Name: "limit", Description: tmmsgs.APIParamLimit},
+			{Name: "after", Description: tmmsgs.APIParamAfter},
```
nguyer, line 37: Are these applicable for a POST?

nguyer (changes requested):
There are a few minor things we should probably fix up (mostly around descriptions) but this all makes sense to me. Looks good!

internal/persistence/leveldb_persistence.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (p *leveldbPersistence) streamKey(streamID *fftypes.UUID) []byte {
+	return []byte(fmt.Sprintf("eventstreams/%s", streamID))
+}
+
+func (p *leveldbPersistence) listenerKey(listenerID *fftypes.UUID) []byte {
+	return []byte(fmt.Sprintf("listeners/%s", listenerID))
+}
+
+func (p *leveldbPersistence) writeJSON(ctx context.Context, key []byte, target interface{}) error {
```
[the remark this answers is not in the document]
peterbroadhurst: Picked `value`

internal/persistence/leveldb_persistence.go
```diff
@@ -0,0 +1,228 @@
...
+}
+
+func (p *leveldbPersistence) GetCheckpoint(ctx context.Context, streamID *fftypes.UUID) (cp *apitypes.EventStreamCheckpoint, err error) {
+	err = p.readJSON(ctx, p.checkpointKey(streamID), &cp)
+	return cp, err
+}
+
+func (p *leveldbPersistence) DeleteCheckpoint(ctx context.Context, streamID *fftypes.UUID) error {
+	return p.deleteKeys(ctx, p.checkpointKey(streamID))
+}
```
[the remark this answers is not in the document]
peterbroadhurst: Great point. Yes, currently it's the responsibility of the caller to map the errors.
However, it's definitely possible (and likely) raw LevelDB errors are working their way through - happy to update that to wrap the errors.

pkg/fftm/manager.go
```diff
@@ -79,16 +84,48 @@ type manager struct {
...
+	}
+	return m, nil
+}
+
+func newManager(ctx context.Context, connector ffcapi.API) *manager {
 	m := &manager{
-		connectorAPI:     ffcapi.NewFFCAPIClient(ctx, tmconfig.ConnectorPrefix, ffcapi.Variant(config.GetString(tmconfig.ConnectorVariant))),
-		ffCoreClient:     ffresty.New(ctx, tmconfig.FFCorePrefix),
+		connector:        connector,
+		ffCoreClient:     ffresty.New(ctx, tmconfig.FFCoreConfig),
```
nguyer, line 121: I think we should have a check somewhere in this function to make sure `ffcore.url` is set. (Or maybe in ffresty?) The default value is `nil` and it will get initialized with an empty string without returning an error. This results in a failure later when code attempts to use the manager at runtime, rather than immediately printing an error telling the user the missing config key at startup.

internal/persistence/leveldb_persistence.go
```diff
@@ -0,0 +1,228 @@
...
+	return l, err
+}
+
+func (p *leveldbPersistence) WriteListener(ctx context.Context, spec *apitypes.Listener) error {
+	return p.writeJSON(ctx, p.listenerKey(spec.ID), spec)
+}
+
+func (p *leveldbPersistence) DeleteListener(ctx context.Context, listenerID *fftypes.UUID) error {
+	return p.deleteKeys(ctx, p.listenerKey(listenerID))
+}
```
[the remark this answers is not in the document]
peterbroadhurst: Think the wrapping is all sorted now

pkg/fftm/route__root_command.go
peterbroadhurst: Think returning 202 is fine 👍

pkg/fftm/route__root_command.go
peterbroadhurst: Also added `deployRequest` to to the input schema anyOf definition

pkg/fftm/manager.go
peterbroadhurst: This has been removed now, as all remnants of the code that used the FF Core SPI have been removed (in favor of using the internal LevelDB and emitting WS replies)

nguyer (approved):
Having been keeping up with changes as they are merged or committed to this branch, and having done extensive testing on it, I think this is stable enough for a merge to main now.

PR #10, opened 2022-07-19
Various fixes for fftm to work with core

nguyer:
This PR is a step in the right direction of getting a working evmconnect to talk to FireFly Core. With these changes transactions work, but events are still a work in progress.

These changes require: https://github.com/hyperledger/firefly-common/pull/26


go.mod
```diff
@@ -57,3 +57,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+replace github.com/hyperledger.org/firefly-common => ../firefly-common
```
peterbroadhurst, line 61: Should we check this in, or just have it locally?

internal/confirmations/confirmations.go
```diff
@@ -514,7 +514,7 @@ func (bcm *blockConfirmationManager) processBlock(block *BlockInfo) {
 
 			// The block might appear at any point in the confirmation list
 			expectedParentHash := pending.blockHash
-			expectedBlockNumber := pending.blockNumber + 1
+			expectedBlockNumber := pending.blockNumber
```
peterbroadhurst, line 517: I could do with some context for this one. This change looks wrong on the face of it (and is in code that overall has been used in anger elsewhere), but I might have missed something.

The `+ 1` here is stating the that the first confirmation, will be the block after the thing that's pending.

e.g. if a transaction is in block `1000` then the first confirmation will be block `1001`.

internal/events/eventstream.go
```diff
@@ -132,9 +132,7 @@ func NewEventStream(
 		wsChannels:         wsChannels,
 		retry:              esDefaults.retry,
 		checkpointInterval: config.GetDuration(tmconfig.EventStreamsCheckpointInterval),
-	}
-	if config.GetInt(tmconfig.ConfirmationsRequired) > 0 {
-		es.confirmations = confirmations.NewBlockConfirmationManager(esCtx, connector)
+		confirmations:      confirmations.NewBlockConfirmationManager(esCtx, connector),
```
peterbroadhurst, line 135: I don't remember us looking to remove this here. I know we discussed doing that the `Manager` would use confirmation manager even if `ConfirmationsRequired`, because the receipt checking is in the confirmations manager for that path. But here I thought we were going to leave the bypass in place.

pkg/fftm/manager.go
```diff
@@ -223,23 +230,26 @@ func (m *manager) fullScan() error {
...
+				log.L(m.ctx).Debugf("Finished reading all operations for namespace '%s' - %d read, %d added", ns, read, added)
+				break
+			}
+			lastOp = ops[len(ops)-1]
+			read += len(ops)
+			for _, op := range ops {
+				added++
+				m.trackIfManaged(op)
+			}
+			page++
```
peterbroadhurst, line 249: Does `page` need resetting to `0` in the outer `ns` loop?

go.mod
nguyer: Probably just locally. I can remove this. I wasn't sure how checks in GitHub would work while we're in such a state of flux - though obviously this particular line will never work there, so I'll remove it.

internal/confirmations/confirmations.go
nguyer: Yeah I'm not sure if this is "correct" or not. My observation with mining on demand was that a transaction wouldn't be marked as completed in FireFly until another block was mined. This seemed to be an "off by one" issue, and this did make it work, though like I said I'm not sure it's the "right" solution.

internal/events/eventstream.go
nguyer: With the check here, if confirmations was set to zero, it would never create the confirmation manager at all. This change makes it always create a confirmation manager, regardless of what confirmations is set to, which is what I thought we discussed.

pkg/fftm/manager.go
nguyer: Yes. Good catch, thanks.

nguyer:
Okay I think the issues we discussed are resolved now. There are still issues (not new issues) with receipts, but transactions do go all the way through, and I see them confirmed on other FireFly nodes that are still using Ethconnect. So that's progress.

peterbroadhurst (approved):
Super, thanks @nguyer  - looks great

PR #12, opened 2022-07-22
Move to internal persistence of the in-flight transactions

peterbroadhurst:
This is one big change on the path to making FFTM/EVMConnect work near-identically to EthConnect.

This swaps from using the FF Core `operations` SPI for keeping track of transactions, to using the built-in LevelDB Persistence layer (previous PR #9 laid the foundations there).

This is not fully UT covered, or e2e tested, but it's a big change so I'd like to merge it with your feature branch work @nguyer sooner than later.

Key things in this PR are:
- Ripping out all the code that talks to FF Core over SPI or WebSockets
- Rebuilding the nonce management logic, based on the LevelDB state APIs
   - This includes a model for what "wins" between local FFTM state vs. the node's own TX Pool when assigning a next nonce

Key functional outstanding items include:
- WebSocket events for `replies` - which really are more `updates` than replies now, as you might get many for an individual transaction as it makes its way to the chain. Need to reconcile that with FF Core in the next PR.
- Get by ID for transactions on REST API
- Delete by ID for transactions on REST API (feel this is going to be needed for recovery scenarios at some point)


nguyer (approved):
I reviewed these changes and tested them Friday. Transaction submission still works great. Operations in FireFly are stuck in "pending" state, but I think that's to be expected given these changes, and the fact that they now need to be sent over the WebSocket. I think this is safe to merge, and I'll pick up getting receipts working from here. Thanks!

PR #13, opened 2022-07-29
Events fixes

nguyer:
This PR adds some fixes for event and receipt delivery over websockets.

### Tested and working:
🟢 Event delivery with 0 required confirmations and on demand mining (standard E2E test configuration)
🟢 Event delivery with 2 required confirmations and 2 second block period

### Tested and sometimes working:
🟡 Receipt delivery over WebSocket with 0 required confirmations and on demand mining (standard E2E test configuration). More details below.

### Tested and not working:
🔴 Receipt delivery over WebSocket with 2 required confirmations and 2 second block period. More details below

### Open questions
- Where is the right place in the code to send the WebSocket message when we have a transaction receipt to deliver? It's currently in the `receiptCallback` function in `policyLoop.go` but it's part of a `manager`. Feels like it might be the right code, but maybe the wrong file?
- I don't think I have the code quite right to handle when a transaction is "done" and confirmed. There are several states a transaction can be in, and I don't think that concept used to exist in the previous ethconnect receipt model.

### Known issues
- When required confirmations is set to `0` and using on demand mining, occasionally (seems like when I send several transactions in quick succession) it will log something like:
```
[2022-07-28T12:58:04.695-04:00] DEBUG evmconnect: Receipt for transaction 0xeb84df74d0354b91b696cf4f964e58fbf6b9f25e0df0a95186b5c1fda83bd616 not yet available
```
There is a comment at the end of that function which says:
```
// No need to keep polling - either we now have a receipt, or normal block header monitoring will pick this one up
```
However it appears that the receipt is not picked up by the normal block header monitoring, or for some reason its receipt callback function is not called. This is the yellow bullet point above.

- When required confirmations is set to `2` and using a 2 second block period, it looks like the receipt callback is never called at this point.

### DIFF
--- a/internal/confirmations/confirmations.go
+++ b/internal/confirmations/confirmations.go
@@ -423,10 +423,16 @@ func (bcm *blockConfirmationManager) checkReceipt(pending *pendingItem) {
 		if pending.receiptCallback != nil {
 			pending.receiptCallback(res)
 		}
-		// Need to walk the chain for this new receipt
-		if err = bcm.walkChainForItem(pending); err != nil {
-			log.L(bcm.ctx).Debugf("Failed to walk chain for transaction %s: %s", pending.transactionHash, err)
-			return
+
+		if bcm.requiredConfirmations == 0 {
+			delete(bcm.pending, pending.getKey())
+			bcm.dispatchConfirmed(pending)
+		} else {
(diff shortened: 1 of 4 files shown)


peterbroadhurst (approved):
Merging this in and working forwards from there

PR #17, opened 2022-08-02
Restructure TX input for consistency with ethconnect and idempotence

peterbroadhurst:
... also removes the last vestige of a dependency on FF Core 🪶 

Updates the reply processing to emit all updates - note that FF core will need to process a new `header.type` of `TransactionUpdate` (as well as `TransactionFailure` and `TransactionSuccess`).

I've merged #15 into this branch


nguyer:
> note that FF core will need to process a new `header.type` of `TransactionUpdate` (as well as `TransactionFailure` and `TransactionSuccess`).

Does this mean that existing apps (old versions of FireFly included) will not work with this connector framework / evmconnect without modification?

peterbroadhurst:
Yeah - I believe that's going to be true with EVMConnect as I see things evolving here.

What I hope can be true is:
- Applications built against EthConnect can be updated to use EVMConnect with very minimal changes
- Once updated applications can use a subset of functionality interchangeably between EthConnect and EVMConnect
  - This subset is deliberately being targeted at the APIs used by FireFly Core

pkg/apitypes/managed_tx.go
```diff
@@ -57,20 +57,43 @@ type ManagedTXError struct {
...
+
+const (
+	TransactionUpdate        ReplyType = "TransactionUpdate"
+	TransactionUpdateSuccess ReplyType = "TransactionSuccess"
+	TransactionUpdateFailure ReplyType = "TransactionFailure"
+)
+
+type ReplyHeaders struct {
+	RequestID string `json:"requestId"`
+	Type      ReplyType
```
nguyer, line 90: I'm not sure if FireFly will care about this, but it might help

```suggestion
	Type      ReplyType `json:"type"`
```

nguyer (approved):
This looks great, thanks! One minor suggestion for a JSON tag, but other than that, looks good to merge.

PR #20, opened 2022-08-19
Fix event stream topic selection

nguyer:
The event stream name was being used to select which websocket connection to send an event. If the event stream name was different from the actual websocket topic, this didn't work. This PR fixes that.

### DIFF
--- a/internal/events/eventstream.go
+++ b/internal/events/eventstream.go
@@ -164,7 +164,7 @@ func (es *eventStream) initAction(startedState *startedStreamState) {
 	case apitypes.EventStreamTypeWebhook:
 		startedState.action = newWebhookAction(ctx, es.spec.Webhook).attemptBatch
 	case apitypes.EventStreamTypeWebSocket:
-		startedState.action = newWebSocketAction(es.wsChannels, es.spec.WebSocket, *es.spec.Name).attemptBatch
+		startedState.action = newWebSocketAction(es.wsChannels, es.spec.WebSocket, *es.spec.WebSocket.Topic).attemptBatch
 	default:
 		// mergeValidateEsConfig always be called previous to this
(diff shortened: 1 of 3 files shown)


internal/events/eventstream.go
```diff
@@ -164,7 +164,7 @@ func (es *eventStream) initAction(startedState *startedStreamState) {
 	case apitypes.EventStreamTypeWebhook:
 		startedState.action = newWebhookAction(ctx, es.spec.Webhook).attemptBatch
 	case apitypes.EventStreamTypeWebSocket:
-		startedState.action = newWebSocketAction(es.wsChannels, es.spec.WebSocket, *es.spec.Name).attemptBatch
+		startedState.action = newWebSocketAction(es.wsChannels, es.spec.WebSocket, *es.spec.WebSocket.Topic).attemptBatch
```
peterbroadhurst, line 167: What is the migration impact of this one? Do we need to edit the Topic of existing pre-v1.1 event streams on startup (before this) to ensure migration works with EthConnect v1.0 environments?

peterbroadhurst (commented):
Question on there - @awrichar might know the answer to this one

peterbroadhurst:
ok - sorry, misunderstood what code this was in. Going to take a deeper look

peterbroadhurst (approved):
So the sequence of how we got here is:
- EVM Connect removed `websocket.topic` because:
   - It was not ensured to be unique
   - I did not have any real value vs. using `name` for the same purpose
   - If you used the same `topic` for two streams, the behavior was undefined
- Due to compatibility convenience with ERC-20/ERC-721, it was re-instated
  - This fix missed making it default to `name` (I've added that here)
  - This fix missed the problem described in this PR

So I've updated this to do what we discussed was the re-instatement of `websocket.topic`, but I am still wondering whether it would be easier to simply make the token connectors migrate.

PR #23, opened 2022-08-22
Add checkpoint/catchup information from FFCAPI to listener GET API

### LINKED ISSUE #22: Add block high watermark information to listener/subscription APIs
Discussed with @shorsher and @nguyer that it would be really helpful to be able to know where an individual EVMConnect listener is up to, if it's catching up from an old block height.

Then FireFly Core could expose this extra information under the listener.

peterbroadhurst:
Resolves #22

- Calls the existing `EventListenerHWM` endpoint in the FFCAPI and mixes that into the API response for the listener
- Adds a `catchup` flag to that API, that allows an FFCAPI implementation to mark a stream as in catchup mode

### DIFF
--- a/pkg/apitypes/api_types.go
+++ b/pkg/apitypes/api_types.go
@@ -117,6 +117,11 @@ type Listener struct {
 	FromBlock        *string           `ffstruct:"listener" json:"fromBlock,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


nguyer (approved):
Looks good. Thanks!

PR #25, opened 2022-08-22
Add delete functionality with policy engine involvement

peterbroadhurst:
See https://github.com/hyperledger/firefly-transaction-manager/issues/24#issuecomment-1222821262

- New `DELETE` `/transactions/{transactionId}` API
- Adds new `DeleteRequested` field to the `ManagedTX` structure
- Changes `updated` boolean return on policy engine `Execute()` to an enum:
   - `UpdateYes` - same as current `true`
   - `UpdateNo` - same as current `false`
   - `UpdateDelete` - the transaction will be deleted from the persistence, with one last event emitted
- Invokes the policy engine once in-line with the API request
  - Returns `200` if that results in immediate removal
  - Returns `202` if the removal is not immediate, but the policy engine does not return an error
  - Returns an error if that run of the policy engine fails
- The policy engine executes this request sequentially with it's other processing


pkg/fftm/route_delete_transaction.go
```diff
@@ -0,0 +1,45 @@
...
+		Path:   "/transactions/{transactionId}",
+		Method: http.MethodDelete,
+		PathParams: []*ffapi.PathParam{
+			{Name: "transactionId", Description: tmmsgs.APIParamTransactionID},
+		},
+		QueryParams:     nil,
+		Description:     tmmsgs.APIEndpointDeleteTransaction,
+		JSONInputValue:  nil,
+		JSONOutputValue: func() interface{} { return &apitypes.ManagedTX{} },
+		JSONOutputCodes: []int{http.StatusAccepted},
```
nguyer, line 39: Do we also need `200` here on this list?

nguyer (approved):
One quick question about status codes, but otherwise looks good!

PR #26, opened 2022-08-23
Remove the websocket.topic for simplicity and to avoid undefined behavior

peterbroadhurst:
The original intention during EVMConnect was to remove this field, because the behavior around it in EthConnect was confusing.
- It had to be unique to each Event Stream
- We didn't enforce it to be unique
- If it wasn't unique, the behavior was undefined

However, we thought that compatibility/migration in FireFly Core and the ERC-20/ERC-721 and ERC-1155 token connectors would be a pain. So we put it back in...

We've since worked through all that in the tokens thanks to @awrichar in the following PRs, so this PR pulls it back out again:
- FireFly Core Ethereum Connector: ???
- ERC-1155: https://github.com/hyperledger/firefly-tokens-erc1155/pull/94
- ERC-20/ERC-721: https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/83

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,6 +9,7 @@
     "badurl",
     "blocklistener",
     "ccache",
+    "Compat",
     "confirmationsmocks",
     "dataexchange",
     "Debugf",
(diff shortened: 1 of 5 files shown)


nguyer (approved):
Looks good to me.

awrichar:
For the record, the change in the Ethereum plugin in core was here: https://github.com/hyperledger/firefly/pull/999

PR #29, opened 2022-08-26
Integrating sprig into gasOracle.Template

hfuss:
http://masterminds.github.io/sprig/ is a common supplemental library for Go templating that has handy functions for ints and floats.

For oracle templates where we need to convert from gwei to wei, and /or convert from float to int, these utility functions can be very useful. This also provides us with a good starting point of examples in the event we ever need to write our own Go template functions such as converting from hex to bigints, etc.

Thanks @peterbroadhurst for doing the hard part of getting everything in place.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,9 @@ require (
 )
 
 require (
+	github.com/Masterminds/goutils v1.1.1 // indirect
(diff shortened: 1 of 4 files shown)


peterbroadhurst (approved):
Thanks @hfuss and @jimthematrix

PR #32, opened 2022-08-29
Pass options.signer for EthCompat mode

peterbroadhurst:
EVMConnect will separate out the option `options.signer: true` from the `options.methods: []` list of methods to match, so you can just request `signer`. Also so that even if no methods match we still set the signer.

This means a tweak to the EthCompat mode


nguyer (approved):
Reviewed this with the other PR and it looks good. Thanks.

PR #33, opened 2022-09-02
GET /status/* APIs for Liveness and Readiness Probes

hfuss:
Adding endpoints for use within K8s for liveness / readiness probes, as well as for upstream connector services like tokens. See https://github.com/hyperledger/firefly-tokens-erc20-erc721/pull/85 where we noticed we are missing these endpoints that Ethconnect did have.

TODO 
- [x] unit tests


pkg/ffcapi/is_ready.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ffcapi
+
+import "github.com/hyperledger/firefly-common/pkg/fftypes"
+
+type ReadyResponse struct {
+	Ready             bool             `json:"ready"`
+	DownstreamDetails *fftypes.JSONAny `json:"downstreamDetails"` // The data output from the method call - can be array or object structure
```
hfuss, line 23: Idea behind this it can include what the downstream is, how it might be erroring if its down, etc.

peterbroadhurst (approved):

peterbroadhurst:
This is great 👍

PR #39, opened 2022-10-07
Add 'fixed' to the enum options for gas oracle mode

chrisbygrave:
- https://github.com/hyperledger/firefly-evmconnect/pull/36

Signed-off-by: Chris Bygrave <chris.bygrave@gmail.com>


peterbroadhurst (commented):
Could you do a `go get github.com/hyperledger/firefly-common@v1.1.3`, to pull in other docs changes.

peterbroadhurst (changes requested):
... also you'll need to do `make reference` so the checked in `config.md` matches the latest (the PR build enforces that - so is failing)

chrisbygrave:
Looks like the doc generation in firefly-common has changed slightly between the versions - a header parameter was introduced so I've updated accordingly.

peterbroadhurst (approved):

pkg/fftm/config_docs_generate_test.go
```diff
@@ -29,12 +29,31 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const configDocHeader = `---
```
peterbroadhurst, line 32: I note you don't actually need this. It was inherited from `firefly-core` where the docs were being generated into Jekyll format, and looks ugly when you view in Github.

The new interface allowing `GenerateConfigMarkdown` to accept a config header, means code (like here) can customize that header of it wants to.

peterbroadhurst: ... where the simplest form of customization is just not to have a header at all

peterbroadhurst: You can see the ugly version here: https://github.com/hyperledger/firefly-transaction-manager/blob/main/config.md

peterbroadhurst (changes requested):
The validation check test also needs updating, sorry @chrisbygrave 

```
# github.com/hyperledger/firefly-transaction-manager/pkg/fftm [github.com/hyperledger/firefly-transaction-manager/pkg/fftm.test]
Error: pkg/fftm/config_docs_test.go:36:55: not enough arguments in call to config.GenerateConfigMarkdown
	have (context.Context, []string)
	want (context.Context, string, []string)
FAIL	github.com/hyperledger/firefly-transaction-manager/pkg/fftm [build failed]
```

pkg/fftm/config_docs_generate_test.go
chrisbygrave: I went like-for-like - happy to change it to simply:

### Configuration reference

in both `config_docs_test.go` and `config_docs_generate_test.go` if GitHub is the only format we're going to use this in.

peterbroadhurst: Yeah - this only goes to Github

peterbroadhurst (approved):

PR #40, opened 2022-10-17
adding a flag to support pre signed transactions

Chengxuan:
Adding a boolean flag in the Managed Transaction to indicate whether the transaction is pre-signed. Therefore, the connector can use that flag to call `sendRawTransaction` .

### DIFF
--- a/pkg/apitypes/managed_tx.go
+++ b/pkg/apitypes/managed_tx.go
@@ -76,6 +76,7 @@ type ManagedTX struct {
 	ErrorMessage       string                             `json:"errorMessage,omitempty"`
 	ErrorHistory       []*ManagedTXError                  `json:"errorHistory"`
 	Confirmations      []confirmations.BlockInfo          `json:"confirmations,omitempty"`
+	PreSigned          bool                               `json:"preSigned,omitempty"`
 }
 
 type ReplyType string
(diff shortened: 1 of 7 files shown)


peterbroadhurst (changes requested):
If I read this right @Chengxuan, the signing step is coupled with the `TransactionPrepare` request.
That is the step in FFCAPI that happens *in-line* with the API call to submit the transaction in, so that a user can get an immediately response that their transaction is valid.

To date this has been only checking the inputs are valid according to the contract - not signing. So allowing an FFCAPI implementation to do signing at this point would be a big change to the API contract.

In fact, I realize as I write this that it happens before nonce allocation - so I don't think it's possible for signing to happen in the `TransactionPrepare` phase.

I would have thought that if an Policy Engine implementation wanted to pre-sign transactions before `TransactionSend` then we would just have it as a boolean on the `TransactionSendRequest` (no other changes needed in FFTM at all - it would be the Policy Engine's responsibility to orchestrate the signing using some other means than FFCAPI, if that's the intent).

Chengxuan:
@peterbroadhurst Good point. I've updated the PR to remove the code changes for TransactionPrepare stage.

peterbroadhurst (approved):
Thanks @Chengxuan - this reduced change does look like all that's needed to me 👍

PR #42, opened 2022-11-22
Add Errors to transaction input to support custom errors

jimthematrix:
See descriptions in https://github.com/hyperledger/firefly-evmconnect/pull/42

### DIFF
--- a/pkg/ffcapi/api.go
+++ b/pkg/ffcapi/api.go
@@ -197,6 +197,7 @@ type TransactionInput struct {
 	TransactionHeaders
 	Method *fftypes.JSONAny   `json:"method"`
 	Params []*fftypes.JSONAny `json:"params"`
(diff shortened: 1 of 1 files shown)


pkg/ffcapi/api.go
```diff
@@ -197,6 +197,7 @@ type TransactionInput struct {
 	TransactionHeaders
 	Method *fftypes.JSONAny   `json:"method"`
 	Params []*fftypes.JSONAny `json:"params"`
+	Errors []*fftypes.JSONAny `json:"errors"`
```
peterbroadhurst, line 200: Noting that in `ContractDeployPrepareRequest` we provide the full ABI, so those are already covered there 👍

peterbroadhurst (approved):

PR #43, opened 2022-11-27
Update FF every time there is an error, or interesting change in state

peterbroadhurst:
Note this is an interim change, and we should fully implement https://github.com/hyperledger/firefly/issues/1108 such that only a transition of the _state_ to `Error` would result in a WebSocket update (not simply us writing an error to our local state from a single execution of the polling loop).

This PR:
1. Fixes the problem that we didn't propagate error updates at all
2. Introduces a more complete history list, that includes when we submitted the transaction, when we got a receipt etc.


matthew1001:
https://github.com/hyperledger/firefly/issues/1108 is largely complete but with a couple of questions if you get a moment to look at them @peterbroadhurst.

peterbroadhurst:
> Looks good. Are there any concerns about compatibility with existing apps with this change?

Great quetion.
I assert it's extremely unlikely that any apps will have depended on the history within the `errorHistory` array. The `errorMessage` field has not changed.

PR #45, opened 2022-12-05
Reduce the size of websocket updates for transactions.

matthew1001:
Subscribers can use /transactions/{txid} to provide full detail about a transaction if required.

Delivered as part of closing https://github.com/hyperledger/firefly/issues/1108

Signed-off-by: Matthew Whitehead <matthew.whitehead@kaleido.io>

### DIFF
--- a/pkg/ffcapi/api.go
+++ b/pkg/ffcapi/api.go
@@ -134,6 +134,13 @@ func (eid *EventID) ProtocolID() string {
 	return fmt.Sprintf("%.12d/%.6d/%.6d", eid.BlockNumber, eid.TransactionIndex, eid.LogIndex)
 }
 
+func ProtocolIDForReceipt(receipt *TransactionReceiptResponse) string {
+	if receipt == nil {
+		return ""
+	}
+	return fmt.Sprintf("%.12d/%.6d", receipt.BlockNumber.Int(), receipt.TransactionIndex.Int())
+}
+
 // Events array has a natural sort order of the block/txIndex/logIndex
 type Events []*Event
 
--- a/pkg/ffcapi/api.go
+++ b/pkg/ffcapi/api.go
@@ -134,13 +134,6 @@ func (eid *EventID) ProtocolID() string {
 	return fmt.Sprintf("%.12d/%.6d/%.6d", eid.BlockNumber, eid.TransactionIndex, eid.LogIndex)
 }
 
-func ProtocolIDForReceipt(receipt *TransactionReceiptResponse) string {
-	if receipt == nil {
-		return ""
-	}
-	return fmt.Sprintf("%.12d/%.6d", receipt.BlockNumber.Int(), receipt.TransactionIndex.Int())
-}
-
 // Events array has a natural sort order of the block/txIndex/logIndex
 type Events []*Event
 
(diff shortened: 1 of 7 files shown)


pkg/ffcapi/api.go
```diff
@@ -134,6 +134,13 @@ func (eid *EventID) ProtocolID() string {
 	return fmt.Sprintf("%.12d/%.6d/%.6d", eid.BlockNumber, eid.TransactionIndex, eid.LogIndex)
 }
 
+func ProtocolIDForReceipt(receipt *TransactionReceiptResponse) string {
```
peterbroadhurst, line 137: I think this is a duplicate from https://github.com/hyperledger/firefly-transaction-manager/pull/43 - so you'll need an update once that merges.

pkg/fftm/policyloop.go
```diff
@@ -286,22 +286,26 @@ func (m *manager) execPolicy(ctx context.Context, pending *pendingState, syncDel
 }
 
 func (m *manager) sendWSReply(mtx *apitypes.ManagedTX) {
```
peterbroadhurst, line 288: Think this function needs a name change, as it's a conditional processing.
Maybe `checkSendWSReply` or `sendWSReplyIfResolved`

pkg/ffcapi/api.go
matthew1001: Yeah I used the same function from #43 knowing we would be merging around the same time. Happy to do the merge once #43 is in.

pkg/fftm/policyloop.go
matthew1001: Yes agreed. Because I'm not a fan of `do-a-thing-if-a-condition-is-met` function names I've chosen to put the status check outside the `sendWSReply` function call.

pkg/fftm/policyloop.go
```diff
@@ -317,26 +326,32 @@ func (m *manager) execPolicy(ctx context.Context, pending *pendingState, syncDel
...
 	wsr := &apitypes.TransactionUpdateReply{
-		ManagedTX: *mtx,
 		Headers: apitypes.ReplyHeaders{
 			RequestID: mtx.ID,
 		},
+		Status:          mtx.Status,
+		TransactionHash: mtx.TransactionHash,
+	}
+
+	if mtx.Receipt != nil {
```
peterbroadhurst, line 344: This feels a little unnecessary to me. It extends the difference between notifications and normal just `GET` `/transactions` output, which are really different for nothing but historical reasons (EthConnect).

However, happy to leave that with you for a final decision @matthew1001

peterbroadhurst (approved):
One comment for you to consider, with approval

pkg/fftm/policyloop.go
peterbroadhurst: (specifically I mean the pulling one field out of `receipt` up to the top-level, vs. the app being notified accessing it at `receipt.protocolId`)

matthew1001: I think it makes sense to me, given it's expect in the notification. Suggest leaving for now and can go back round the loop in a future PR?

PR #46, opened 2022-12-08
Add AddressBalance() to FFCAPI and a new REST API route to call it

matthew1001:
The aim of this PR is to add an address balance API call to the FFCAPI interface. This makes it possible for users of of the FFCAPI (such as FireFly Transaction Manager) to query and optionally surface via their own APIs.

Under this PR I have introduced a new REST API `/gastoken/balances/{address}` which calls `AddressBalance()` for the the configured blockchain connector. The use of `gastoken` in the API path is intended to be clear that it is unrelated to token balances.

Signed-off-by: Matthew Whitehead <matthew.whitehead@kaleido.io>


internal/tmmsgs/en_api_descriptions.go
```diff
@@ -51,6 +51,7 @@ var (
 	APIEndpointPostEventStreamListenerReset = ffm("api.endpoints.post.eventstream.listener.reset", "Reset an event stream listener, to redeliver all events since the specified block")
 	APIEndpointPatchEventStreamListener     = ffm("api.endpoints.patch.eventstream.listener", "Update event stream listener")
 	APIEndpointDeleteEventStreamListener    = ffm("api.endpoints.delete.eventstream.listener", "Delete event stream listener")
+	APIEndpointGetAddressBalance            = ffm("api.endpoints.get.address.balance", "Get balance for a signer address")
```
peterbroadhurst, line 54:
```suggestion
	APIEndpointGetAddressBalance            = ffm("api.endpoints.get.address.balance", "Get gas token balance for a signer address")
```

internal/tmmsgs/en_api_descriptions.go
```diff
@@ -60,4 +61,5 @@ var (
 	APIParamTXSigner      = ffm("api.params.txSigner", "Return only transactions for a specific signing address, in reverse nonce order")
 	APIParamTXPending     = ffm("api.params.txPending", "Return only pending transactions, in reverse submission sequence (a 'sequenceId' is assigned to each transaction to determine its sequence")
 	APIParamSortDirection = ffm("api.params.sortDirection", "Sort direction: 'asc'/'ascending' or 'desc'/'descending'")
+	APIParamSignerAddress = ffm("api.params.signerAddress", "A signing address, for example to get the balance for")
```
peterbroadhurst, line 64:
```
	APIParamSignerAddress = ffm("api.params.signerAddress", "A signing address, for example to get the gas token balance for")
```

pkg/fftm/route_get_address_balance.go
```diff
@@ -0,0 +1,44 @@
...
+
+var getAddressBalance = func(m *manager) *ffapi.Route {
+	return &ffapi.Route{
+		Name:   "getBalance",
+		Path:   "/gastoken/balances/{address}",
+		Method: http.MethodGet,
+		PathParams: []*ffapi.PathParam{
+			{Name: "address", Description: tmmsgs.APIParamSignerAddress},
+		},
+		QueryParams:     nil,
```
peterbroadhurst, line 35: Should there not be a query param for `blocktag` here?

peterbroadhurst (commented):
Two minor suggestions just to clarify which balance is being queried (e.g. it's not a coin, but the underlying native gas token of the chain), and a question on whether the `blockTag` field that's added to the FFCAPI should be included on the REST API.

internal/tmmsgs/en_api_descriptions.go
matthew1001: Changed.

pkg/fftm/route_get_address_balance.go
```diff
@@ -0,0 +1,46 @@
...
+var getAddressBalance = func(m *manager) *ffapi.Route {
+	return &ffapi.Route{
+		Name:   "getBalance",
+		Path:   "/gastoken/balances/{address}",
+		Method: http.MethodGet,
+		PathParams: []*ffapi.PathParam{
+			{Name: "address", Description: tmmsgs.APIParamSignerAddress},
+		},
+		QueryParams: []*ffapi.QueryParam{
+			{Name: "blockTag", Description: tmmsgs.APIParamBlockTag},
```
peterbroadhurst, line 36: Tiny point, but while these are case insensitive we've previously tried to make the docs the lower-case value.

peterbroadhurst (approved):
One single-character change suggestion with approval

PR #54, opened 2023-01-17
Bring FFTM back to 100% test coverage

matthew1001:
Signed-off-by: Matthew Whitehead <matthew1001@gmail.com>

When I added prometheus metrics to FFTM I didn't maintain 100% test coverage, so this PR addresses that and has us back at 100%.

### DIFF
--- a/pkg/fftm/manager.go
+++ b/pkg/fftm/manager.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2023 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -105,10 +105,6 @@ type manager struct {
 func InitConfig() {
 	tmconfig.Reset()
 	events.InitDefaults()
-
-	if config.GetBool(tmconfig.MetricsEnabled) {
-		metrics.Registry()
-	}
 }
 
 func NewManager(ctx context.Context, connector ffcapi.API) (Manager, error) {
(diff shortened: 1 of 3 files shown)


pkg/fftm/manager.go
```diff
@@ -105,10 +105,6 @@ type manager struct {
 func InitConfig() {
 	tmconfig.Reset()
 	events.InitDefaults()
-
-	if config.GetBool(tmconfig.MetricsEnabled) {
-		metrics.Registry()
```
peterbroadhurst, line 110: Won't removing this disable metrics?

peterbroadhurst (commented):
One question on this one @matthew1001

pkg/fftm/manager.go
matthew1001: I re-tested before and after and think I came to the conclusion this was extraneous. `tmconfig.Reset()` sets `MetricsEnabled` to `false` (see https://github.com/hyperledger/firefly-transaction-manager/blob/574b5c73105016c8b6bcd24ff5fefb5858bea7b3/pkg/fftm/manager.go#L106) but now I'm questioning it so let me check.

matthew1001: So looking at the code and re-re-testing with this commit in place, I think this _was_ extraneous code. I think

https://github.com/hyperledger/firefly-transaction-manager/blob/574b5c73105016c8b6bcd24ff5fefb5858bea7b3/pkg/fftm/manager.go#L173

performs the same registry initialisation that was being performed here.

peterbroadhurst: Super, thanks for checking that 👍

peterbroadhurst (approved):

PR #56, opened 2023-01-17
Add chain gas price API

matthew1001:
Closes https://github.com/hyperledger/firefly-transaction-manager/issues/55

Signed-off-by: Matthew Whitehead <matthew1001@gmail.com>

### DIFF
--- a/pkg/fftm/gas_management.go
+++ b/pkg/fftm/gas_management.go
@@ -1,0 +1,37 @@
+// Copyright © 2023 Kaleido, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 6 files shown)


pkg/fftm/gas_management.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"context"
+
+	"github.com/hyperledger/firefly-common/pkg/log"
+	"github.com/hyperledger/firefly-transaction-manager/pkg/apitypes"
+	"github.com/hyperledger/firefly-transaction-manager/pkg/ffcapi"
+)
+
+func (m *manager) getLiveGasPrice(ctx context.Context) (resp *apitypes.LiveGasPrice, err error) {
```
peterbroadhurst, line 27: Just wondering why we embed `ffcapi. GasPriceEstimateResponse ` in a wrapper `apitypes. LiveGasPrice` rather than just using `GasPriceEstimateResponse` directly?

peterbroadhurst (approved):
Looks great. Approved - one comment to consider before merging.

pkg/fftm/gas_management.go
matthew1001: Yeah we seem to have a convention of wrapping `ffacpi` structures in an `apitype` structure when the API call involves a call down into the connector. So e.g. `getLiveStatus()` gets an `ffcapi.LiveResponse` structure back from the connector and wraps it in an `apitypes.LiveStatus` structure. I suspect it's a difference between APIs that call down into the `ffcapi` endpoint and those which generate their responses from FFTM itself. If it's OK I'll merge for now but if we think there's an unnecessary level of abstraction across some of the APIs I'd be happy to go through them all and remove the wrapper layer?

PR #58, opened 2023-01-27
Suggestions on TODOs and comments/naming for TX History API

peterbroadhurst:
@matthew1001 - this is the promised PR review / TODO help from PR #51, covering my high level comments.

I'm afraid I didn't get to looking at the reason you're seeing out-of-order serialization on some paths.

The items I've tried to provide here are:
- Simplifying config to one parameter
    - I believe only one array needs a limit, the others are limited by the list of different types
- Passing config to the TX History API
    - I refactored it out to a separate `txhistory.Manager` instanced, and added that into what's passed to the Policy Engine
    - I called this new structure `policyengine.ToolkitAPI` which I think is in-line with the evolving architecture
- Comments / code naming / removing fields
    - I've added code comments with my understanding of the intent of the different sections, both to validate my current understanding, and to aid future developers of policy engines.


matthew1001:
Thanks for the updates/suggestions. To close the circle on discussions...

1. Re. config, yes simplifying down from 3 config parameters makes it's less cumbersome. I wondered with the new changes if `mtx.HistorySummary` and `currentSubStatus.Actions` can grow indefinitely if a policy engine kept adding new types of sub-status or action?
2. Yeah I'd preferred to go with the idea that you call sub-status functions on a transaction itself rather than call sub-status functions on another entity and pass in the transaction, but this did seem to make passing config through more tricky. I had pondered moving the sub-status functions into the `manager` but opted not to. Think your alternative approach is perfectly valid though and seems to solve the config issue.
3. Re comments etc. yes certainly needed to clarify functions/behaviours more than I had, thanks for adding.

peterbroadhurst:
> if mtx.HistorySummary and currentSubStatus.Actions can grow indefinitely if a policy engine kept adding new types of sub-status or action?

I think this would be considered a code bug in the implementation of the policy engine though right? e.g. not following the spec/spirit of the API.

matthew1001:
> 

Agreed, I think my original approach of making each list configurable was overkill to protect against a mis-implementation of a policy engine but agree that we don't need it.

matthew1001 (approved):
Thanks for the input, think the refactoring makes sense, as does removes the extraneous config points for the different lists

PR #74, opened 2023-03-30
Fixing support for legacy configs

Chengxuan:
for https://github.com/hyperledger/firefly/issues/1244

Still need to support the old configurations in which policyengine.simple is provided.


Chengxuan:
Here is some context if that’s helpful. The old configuration was under the name `policy` and had a default `policy.name` set to `simple`.

After the refactor, the new configuration is called `transactions.handler` . Therefore, to ensure the old logic still works, there is a piece of logic to handling the old policy structure. However, there is a bug in the logic, so it never worked. The PR is to fix the logic by checking whether the new `transactions.handler.name` is set, if it’s not set, then fall back to use the `policy.name: simple` as the configuration.

old configuration to support:
image

new configuration to move into for the next release:
image

pkg/txhandler/registry/registry.go
```diff
@@ -45,14 +45,7 @@ type Factory interface {
...
-	// check whether a policy engine name is provided
-	if config.GetString(tmconfig.DeprecatedPolicyEngineName) != "" {
-		factory.InitConfig(tmconfig.DeprecatedPolicyEngineBaseConfig.SubSection(name))
-	} else {
-		// if not, use the new transaction handler configurations
-		factory.InitConfig(tmconfig.TransactionHandlerBaseConfig.SubSection(name))
-
-	}
+	// init the new transaction handler configurations
+	factory.InitConfig(tmconfig.TransactionHandlerBaseConfig.SubSection(name))
```
awrichar, line 49: Feels slightly odd that one subsection is passed in as an arg, and the other subsection is hard-coded inside the InitConfig method. Maybe both should be passed in _or_ both should be hardcoded?

awrichar (approved):

PR #78, opened 2023-04-27
Update firefly-common

peterbroadhurst:
So that we can pull https://github.com/hyperledger/firefly-signer/releases/tag/v1.1.6 that uses this FF common, into firefly-evmconnect.


peterbroadhurst:
Closing in favor of a PR from @shorsher

PR #82, opened 2023-05-24
Add GasEstimate FFCAPI call

matthew1001:
This allows the `eth_estimateGas` call to be invoked directly.

I haven't included the addition of an API root in this PR. I'm not sure of the use-case for a direct call to `eth_estimateGas` from an API caller, but it's useful for FFTM to have direct access to it.


pkg/ffcapi/api.go
```diff
@@ -40,6 +41,9 @@ type API interface {
 	// NextNonceForSigner is used when there are no outstanding transactions for a given signing identity, to determine the next nonce to use for submission of a transaction
 	NextNonceForSigner(ctx context.Context, req *NextNonceForSignerRequest) (*NextNonceForSignerResponse, ErrorReason, error)
 
+	// GasEstimate provides an estimate of the gas required for the given transaction
+	GasEstimate(ctx context.Context, req *ethsigner.Transaction) (*GasEstimateResponse, ErrorReason, error)
```
peterbroadhurst, line 45: @matthew1001 - just wondering why this isn't exactly the same pattern as:

https://github.com/hyperledger/firefly-transaction-manager/blob/0e611061a7d2daeeaeb0c821fb1584c457b4ccd2/pkg/ffcapi/transaction_prepare.go#L41-L43

matthew1001: Yeah you're right, it should be. Updated in latest commit.

peterbroadhurst (approved):
👍

PR #84, opened 2023-06-02
Persistence enhancements - including adding PostgreSQL support

peterbroadhurst:
Work in progress - release should be marked v1.3 when complete/merged.

Making as pragmatic as possible change to the persistence mechanism, to support PostgreSQL as an alternative to LevelDB when persisting transactions, checkpoints, subscriptions, and listeners.

Aiming to make the impact on policy engine extensions as minimal as possible, but there are some changes shaping up:
- `TXHistory` moving into persistence, as implementation will be fundamentally different for PSQL to LevelDB
     - Updates to substatus history & actions are discrete operations
     - Can now return `error` and only take the `ID` of the `ManagedTX` object (not the full object)
     - History limits only likely to apply to LevelDB (as expensive to implement in SQL)
- Reconciling the `ManagedTX` object into core object fields, and related objects that can be merged in during a `GET` operation for backwards compatibility
     - `History` / `HistorySummary` were the big things
     - `TransactionHeaders` ended up a bit confused, so I've attempted to reconcile it (more info owed by me here)
     - `Receipt` - TBD if this stays separate
      - LevelDB continues to be a single fat JSON payload
- Rather than a single `WriteTransaction` function  with a `new` boolean, moved to `InsertTransaction` and `UpdateTransaction` - with clearer update semantics
      - This avoid re-writing huge JSON blobs to SQL DBs when simply tweaking a status field


peterbroadhurst:
Incorrect target - cancelling

PR #91, opened 2023-07-03
Expose filters for rich query

Chengxuan:
- expose the filters for rich query tables so that they can be used externally.
- ~also updated the event handler callback function signatures to pass through the entire managed transactions so the other information (e.g. "from" field) becomes accessible.~


pkg/fftm/transaction_events_handler.go
```diff
@@ -58,12 +58,12 @@ func (eh *ManagedTransactionEventHandler) HandleEvent(_ context.Context, e apity
...
 				TransactionHash: e.Tx.TransactionHash,
 				Receipt: func(ctx context.Context, receipt *ffcapi.TransactionReceiptResponse) {
-					if err := eh.TxHandler.HandleTransactionReceiptReceived(ctx, e.Tx.ID, receipt); err != nil {
+					if err := eh.TxHandler.HandleTransactionReceiptReceived(ctx, e.Tx, receipt); err != nil {
 						log.L(ctx).Errorf("Receipt for transaction %s at nonce %s / %d - hash: %s was not handled due to %s", e.Tx.ID, e.Tx.TransactionHeaders.From, e.Tx.Nonce.Int64(), e.Tx.TransactionHash, err.Error())
 					}
 				},
 				Confirmations: func(ctx context.Context, notification *apitypes.ConfirmationsNotification) {
-					if err := eh.TxHandler.HandleTransactionConfirmations(ctx, e.Tx.ID, notification); err != nil {
+					if err := eh.TxHandler.HandleTransactionConfirmations(ctx, e.Tx, notification); err != nil {
```
peterbroadhurst: I'm worried about this change.
Passing a pointer to a long term background callback implies an invalid assumption about the use of this callback.

While there might be policy engines for whom this is a safe thing to do, we cannot rely on that in the generic interface.

peterbroadhurst: I reverted this change in my review so this can merge with the important change of providing access to the filters.

peterbroadhurst (approved):
Think we need to work through the second item, so I've added a commit to allow the first to merge.

PR #98, opened 2023-07-26
avoid using internal interface in public pkg

Chengxuan:
In order to using mockery in another golang module, we need to make sure the interfaces are defined in the external packages and consumed by the internal interface, not the other way around. 

Otherwise the consumer will hit problem like the following:
image


pkg/txhandler/txhandler.go
```diff
@@ -19,26 +19,80 @@ package txhandler
...
+type NextNonceCallback func(ctx context.Context, signer string) (uint64, error)
+
+type SortDirection int
+
+const (
+	SortDirectionAscending SortDirection = iota
+	SortDirectionDescending
+)
+
 type TransactionPersistence interface {
```
Chengxuan, line 38: might worth splitting these into different packages. e.g. `pkg/persistence` `pkg/metrics` etc.

peterbroadhurst: Was thinking similar reviewing this.

However, my thought was less about having lots of small packages, but more more that `txhandler` feels like an odd to contain the interfaces of the `toolkit`.
So instead having `toolkit.TransactionPersistence` and `toolkit.NextNonceCallback` and `toolkit.TransactionMetrics` would be better.

However, I'm going to approve and merge what's here as a clear step forwards from the problematic build issue we have now.

peterbroadhurst (approved):
Approved - one comment to consider for a future change.

PR #99, opened 2023-08-02
Fixing a nonce conflict issue when the cached nonce expired.

Chengxuan:
There is a problem when two different batches of transaction inserts happen in parallel, one of them will read the incorrect nonce back from the DB + the nonce callback when the cached nonce expires. 

The fix in the PR is to check the expired cached nonce first to avoid pending transactions that have clashing nonces between them.

### DIFF
--- a/internal/persistence/postgres/transaction_writer.go
+++ b/internal/persistence/postgres/transaction_writer.go
@@ -290,11 +290,12 @@ func (tw *transactionWriter) runBatch(ctx context.Context, b *transactionWriterB
 func (tw *transactionWriter) assignNonces(ctx context.Context, txInsertsByFrom map[string][]*transactionOperation) error {
 	for signer, txs := range txInsertsByFrom {
 		cacheEntry, isCached := tw.nextNonceCache.Get(signer)
+		cacheExpired := false
 		if isCached {
 			timeSinceCached := time.Since(*cacheEntry.cachedTime.Time())
 			if timeSinceCached > tw.p.nonceStateTimeout {
 				log.L(ctx).Infof("Nonce cache expired for signer '%s' after %s", signer, timeSinceCached.String())
-				cacheEntry = nil
+				cacheExpired = true
 			}
 		}
 		for _, op := range txs {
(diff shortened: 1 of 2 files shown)


internal/persistence/postgres/transaction_writer.go
```diff
@@ -306,24 +307,35 @@ func (tw *transactionWriter) assignNonces(ctx context.Context, txInsertsByFrom m
...
+						log.L(ctx).Errorf("Failed to query highest persisted nonce for '%s': %s", signer, err)
+						return err
 					}
+					if len(existingTXs) > 0 {
+						internalNextNonce = existingTXs[0].Nonce.Uint64() + 1
+						log.L(ctx).Tracef("Using the existing nonce from DB %s / %d to compare with the queried next %d for transaction %s", signer, internalNextNonce, nextNonce, op.txInsert.ID)
+					}
+				}
+				if internalNextNonce > nextNonce {
+					log.L(ctx).Infof("Using next nonce %s / %d instead of queried next %d for transaction %s", signer, internalNextNonce+1, nextNonce, op.txInsert.ID)
```
nguyer, line 337: I think this line ends up logging the next, next nonce. Previously the code was `existingNonce+1` but now it is `internalNextNonce+1` which has already been incremented.

nguyer (changes requested):

internal/persistence/postgres/transaction_writer.go
Chengxuan, line 337:
```suggestion
					log.L(ctx).Infof("Using next nonce %s / %d instead of queried next %d for transaction %s", signer, internalNextNonce, nextNonce, op.txInsert.ID)
```

internal/persistence/postgres/transaction_writer.go
```diff
@@ -306,24 +307,35 @@ func (tw *transactionWriter) assignNonces(ctx context.Context, txInsertsByFrom m
...
+					// when there is no cached nonce we need to fetch the highest nonce in our DB
+					filter := persistence.TransactionFilters.NewFilterLimit(ctx, 1).Eq("from", signer).Sort("-nonce")
+					existingTXs, _, err := tw.p.transactions.GetMany(ctx, filter)
+					if err != nil {
+						log.L(ctx).Errorf("Failed to query highest persisted nonce for '%s': %s", signer, err)
+						return err
 					}
+					if len(existingTXs) > 0 {
+						internalNextNonce = existingTXs[0].Nonce.Uint64() + 1
+						log.L(ctx).Tracef("Using the existing nonce from DB %s / %d to compare with the queried next %d for transaction %s", signer, internalNextNonce, nextNonce, op.txInsert.ID)
```
Chengxuan, line 333:
```suggestion
						log.L(ctx).Tracef("Using the next nonce calculated from DB %s / %d to compare with the queried next %d for transaction %s", signer, internalNextNonce, nextNonce, op.txInsert.ID)
```

nguyer (approved):

PR #100, opened 2023-08-30
Update CODEOWNERS

Chengxuan:
Update code owners file to use the github team

@nguyer Not sure what to do with the Doc maintainers, there isn't a github team for them.

image

@nickgaski also lost access to the repo.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,7 +1,3 @@
 # SPDX-License-Identifier: Apache-2.0
 
-# FireFly transaction manager maintainers
 * @hyperledger/firefly-transaction-manager-maintainers
-
-# FireFly Documentation Maintainers
-/docs @peterbroadhurst @nguyer @awrichar @nickgaski
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
-# FireFly Core Maintainers
-* @peterbroadhurst @nguyer @awrichar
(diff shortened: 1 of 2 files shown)


CODEOWNERS
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
-# FireFly Core Maintainers
-* @peterbroadhurst @nguyer @awrichar
+# FireFly transaction manager maintainers
```
nguyer, line 3: Not sure the comment is necessary anymore given that the team name is self-describing

CODEOWNERS
```diff
@@ -1,7 +1,7 @@
...
+* @hyperledger/firefly-transaction-manager-maintainers
 
 # FireFly Documentation Maintainers
-/docs @peterbroadhurst @nguyer @awrichar @nickgaski
\ No newline at end of file
+/docs @peterbroadhurst @nguyer @awrichar @nickgaski
```
nguyer, line 7: I think we can drop this section. We started with the idea of separate docs maintainers early on, but in practice have never used that idea.

nguyer (changes requested):

nguyer:
Do you mind also updating https://github.com/hyperledger/firefly-transaction-manager/blob/main/MAINTAINERS.md

To match the format in https://github.com/hyperledger/firefly/blob/main/MAINTAINERS.md

And add yourself? 🙂

nguyer (approved):

PR #104, opened 2023-12-21
Require policy engines to map failures during submission to idempotency

peterbroadhurst:
See https://github.com/hyperledger/firefly/issues/1435

The implementation here:
- Requires the policy manager implementations to return a new boolean
- Provides a mapping from the `ErrorReason` information that FFCAPI implementations already provide
- Uses a `submissionRejected` boolean on the response back to core (so `false` means it is retryable)
- Updates to latest `firefly-common`


pkg/ffcapi/submission_error.go
```diff
@@ -0,0 +1,40 @@
...
+	// database, so the nonce is not spent, and the transaction is never submitted.
+	SubmissionRejected bool `json:"submissionRejected,omitempty"`
+}
+
+func MapSubmissionRejected(reason ErrorReason) bool {
+	switch reason {
+	case ErrorReasonInvalidInputs,
+		ErrorReasonTransactionReverted,
+		ErrorReasonInsufficientFunds:
+		// These reason codes are considered as rejections of the transaction - see ffcapi.SubmissionError
```
awrichar, line 34: This feels like the most significant part of the change. This list seems good from what I can tell though.

awrichar (approved):

PR #112, opened 2024-03-27
adding index for status field

Chengxuan:
Without an index for the status field, the polling cycle will look up more and more transactions to find "Pending" ones, which causing the performance to decrease:

image


peterbroadhurst (approved):

Chengxuan:
We don't update status very often, only twice in the lifetime for an uninterrupted transaction. So the impact of HOT update is not too significant.

peterbroadhurst:
Great find 🚀

PR #114, opened 2024-04-10
fix null pointer when reading inflight queue

EnriqueL8:
The inflight queue is manipulated by the policy loop. So there is a chance the transactions get cleaned up from the map when other functions are reading it, thus causing the nil pointer exception.

We didn't want to introduce a lock for this as that would slow down the processing of the main policy loop. Therefore, extra checks are added to avoid the exception instead.


Chengxuan (approved):
LGTM

PR #116, opened 2024-04-23
fix: race condition when reading and writing to inflight queue

EnriqueL8:
The policy loop frequently read and writes from the inflight queue, more specifically:
- It will load new transactions into the queue when it has space in `updateInFlightSet` and remove the ones that have completed.
- Remove transactions from the queue when it gets enough confirmations for that transaction

When loading new transactions, the code will take a copy and override the inflight queue with an empty array. If another thread was reading from such array, most often iterating to find a specific inflight tx it wouldn't find it.

This PR alongside with adding extra logging to help diagnose this sort of issues, adds a read/write mutex for the inflight queue and a mutex for each pendingState to be thread safe. Creates read locks when iterating and reading from the queue, write lock in a single place where we remove or load new transactions in queue and a mutex on the pendingState when we read or modify it.

Draft: as running another performance test for the latest commit

### DIFF
--- a/pkg/txhandler/simple/policyloop.go
+++ b/pkg/txhandler/simple/policyloop.go
@@ -107,13 +107,11 @@ func (sth *simpleTransactionHandler) updateInflightSet(ctx context.Context) bool
 	defer sth.inflightRWMux.Unlock()
 
 	oldInflight := sth.inflight
-	// TODOL THIS NEEDS A LOCK!!!
 	sth.inflight = make([]*pendingState, 0, len(oldInflight))
 
 	// Run through removing those that are removed
 	for _, p := range oldInflight {
 		if !p.remove {
-			log.L(sth.ctx).Debugf("Removing TX '%s' from inflight", p.mtx.ID)
 			sth.inflight = append(sth.inflight, p)
 		} else {
 			sth.incTransactionOperationCounter(ctx, p.mtx.Namespace(ctx), "removed")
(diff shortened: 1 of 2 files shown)


pkg/txhandler/simple/policyloop.go
```diff
@@ -152,7 +155,8 @@ func (sth *simpleTransactionHandler) updateInflightSet(ctx context.Context) bool
 		}
 		newLen := len(sth.inflight)
 		if newLen > 0 {
-			log.L(ctx).Debugf("Inflight set updated len=%d head-seq=%s tail-seq=%s old-tail=%s", len(sth.inflight), sth.inflight[0].mtx.SequenceID, sth.inflight[newLen-1].mtx.SequenceID, after)
+			log.L(ctx).Debugf("Inflight set updated with %d additional transactions", len(additional))
+			log.L(ctx).Debugf("Inflight set updated len=%d head-id:%s head-seq=%s tail-id:%s tail-seq=%s old-tail=%s", len(sth.inflight), sth.inflight[0].mtx.ID, sth.inflight[0].mtx.SequenceID, sth.inflight[newLen-1].mtx.ID, sth.inflight[newLen-1].mtx.SequenceID, after)
```
Chengxuan, line 159: need to combine into one

Chengxuan (changes requested):
Minor comments on the loggings. The use of read locks looks good.

pkg/txhandler/simple/policyloop.go
```diff
@@ -170,9 +174,12 @@ func (sth *simpleTransactionHandler) policyLoopCycle(ctx context.Context, inflig
 			return
 		}
 	}
-	// Go through executing the policy engine against them
 
+	sth.inflightRWMux.RLock()
+	defer sth.inflightRWMux.RUnlock()
+	// Go through executing the policy engine against them
 	for _, pending := range sth.inflight {
+		log.L(ctx).Debugf("Executing policy against tx-id=%v", pending.mtx.ID)
```
Chengxuan, line 182:
```suggestion
		log.L(ctx).Tracef("Executing policy against tx-id=%v", pending.mtx.ID)
```

pkg/txhandler/simple/policyloop.go
```diff
@@ -118,6 +120,7 @@ func (sth *simpleTransactionHandler) updateInflightSet(ctx context.Context) bool
 
 	// If we are not at maximum, then query if there are more candidates now
 	spaces := sth.maxInFlight - len(sth.inflight)
+	log.L(sth.ctx).Debugf("Number of spaces left '%v'", spaces)
```
Chengxuan, line 123:
```suggestion
	log.L(sth.ctx).Tracef("Number of spaces left '%v'", spaces)
```

Chengxuan (approved):
LGTM

PR #117, opened 2024-04-26
Pass in action occurred time

### LINKED ISSUE #111: History entries does not reflect the real time the an event occured

Chengxuan:
Resolves https://github.com/hyperledger/firefly-transaction-manager/issues/111

Allow action occurred time to be passed in so the recorded time reflects the time when an action occurred.


Philip-21:
I can see the changes now, i understand it better i should have added the signature `*fftypes.FFTime` in the interface `AddSubStatusAction(ctx context.Context, txID string, subStatus apitypes.TxSubStatus, action apitypes.TxAction, info *fftypes.JSONAny, err *fftypes.JSONAny, actionOccurred *fftypes.FFTime) error
}` instead of removing  it in the main `AddSubStatusAction()`

EnriqueL8 (approved):
This looks great to me! Thanks both

PR #118, opened 2024-05-14
Add synced field to HWM response

chrisbygrave:
Add a synced field to a high water mark response to give connectors a place to record info about whether a listener should be considered in- or out-of-sync

### DIFF
--- a/pkg/ffcapi/event_listener_hwm.go
+++ b/pkg/ffcapi/event_listener_hwm.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2024 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
--- a/pkg/ffcapi/event_listener_hwm.go
+++ b/pkg/ffcapi/event_listener_hwm.go
@@ -27,6 +27,10 @@ type EventListenerHWMRequest struct {
 
 type EventListenerHWMResponse struct {
 	Checkpoint EventListenerCheckpoint `json:"checkpoint"`
-	Catchup    bool                    `json:"catchup,omitempty"` // informational only - informs an operator that the stream is catching up
-	Synced     bool                    `json:"synced,omitempty"`  // derived from the distance of the checkpoint block from chain head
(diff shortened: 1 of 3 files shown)


pkg/ffcapi/event_listener_hwm.go
```diff
@@ -28,4 +28,5 @@ type EventListenerHWMRequest struct {
 type EventListenerHWMResponse struct {
 	Checkpoint EventListenerCheckpoint `json:"checkpoint"`
 	Catchup    bool                    `json:"catchup,omitempty"` // informational only - informs an operator that the stream is catching up
+	Synced     bool                    `json:"synced,omitempty"`  // derived from the distance of the checkpoint block from chain head
```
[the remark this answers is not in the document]
chrisbygrave: That's right. I've added some more comments around the struct to capture what you've written above.

Chengxuan (approved):
LGTM

pkg/ffcapi/event_listener_hwm.go
```diff
@@ -27,5 +27,10 @@ type EventListenerHWMRequest struct {
 
 type EventListenerHWMResponse struct {
 	Checkpoint EventListenerCheckpoint `json:"checkpoint"`
-	Catchup    bool                    `json:"catchup,omitempty"` // informational only - informs an operator that the stream is catching up
+	// Catchup and Synced capture how up-to-date the listener is vs the chain head
+	// If it is a long way back, it will enter a catchup mode
+	// If it is a short way back, it will be considered out-of-sync
+	// If Catchup is true, Synced must be false
+	Catchup bool `json:"catchup,omitempty"` // informational only - informs an operator that the stream is catching up
+	Synced  bool `json:"synced,omitempty"`  // derived from the distance of the checkpoint block from chain head
```
peterbroadhurst, line 35: What is the difference between these two fields?

chrisbygrave: > What is the difference between these two fields?

I'm beginning to think the answer to this question is "Not enough to warrant there being two fields". I think `catchup` is sufficient for my needs... !catchup ~== synced

PR #119, opened 2024-05-19
Update mockery version

alex-semenyuk:
Fix issue on `make mocks`
```
internal error: package "context" without types was imported from "github.com/hyperledger/firefly-transaction-manager/pkg/fftm"
make: *** [mocks--Manager] Error 1
```

Explicit define version instead of latest for mockery as recommended  https://vektra.github.io/mockery/latest/installation/


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hyperledger/firefly-transaction-manager
 
-go 1.21
```
Chengxuan, line 3: We usually do golang version update across Firefly repos and don't normally pin to a specific patch version. e.g., it should be `go 1.22` in this example

@EnriqueL8 is there a plan for the minor version 22 yet?

EnriqueL8: I haven't set a plan yet for it, will keep that in mind - @alex-semenyuk is it strictly needed for this change?

EnriqueL8: @alex-semenyuk ^^^^

alex-semenyuk: @EnriqueL8 sorry for delay, yes, we can move without version update, so updated PR
Btw noticed that there is intermittently failed test https://github.com/hyperledger/firefly-transaction-manager/issues/127

Chengxuan (approved):

PR #120, opened 2024-05-30
Adding metrics for event confirmation and receipt

Chengxuan:
Adding metrics for the following area:
1. Receipt checker - basic counter and duration histogram
2. Confirmation manager
    - notification queue queueing time and process time
    - e2e confirmation time


internal/confirmations/confirmations.go
```diff
@@ -371,6 +385,7 @@ func (bcm *blockConfirmationManager) confirmationsListener() {
 
 		// Mark receipts stale after duration
 		bcm.staleReceiptCheck()
+		log.L(bcm.ctx).Errorf("Confirmation listener processed %d block hashes and %d notifications in %s, trigger type: %s", blockHashCount, notificationCount, time.Since(startTime), triggerType)
```
Chengxuan, line 388: this should be converted into Trace/debug

PR #121, opened 2024-06-05
Adding event processing metrics for Confirmation manager and Receipt checker

Chengxuan:
This PR extends the existing metrics manager to define and emit event processing related metrics for identifying bottlenecks during performance testing. The following metrics have been introduced:
1. Overall processing rate & duration of receipts
2. Overall processing rate & duration of confirmations
3. Processing rate and duration of the following:
    - a notification (asynchronous instructions that the confirmation manager process) joining the queue ( a buffered channel) grouped by notification type (e.g. new transaction, remove transaction, receipt arrived etc)
    - a notification get picked up and processed by the confirmation manager
    - block hash (fetched by event listener) joining the process queue 
    - block hash get processed
4. The receipt checker's process rate grouped by check status and duration
5. Number of block hashes batched for each process loop

The example grafana dashboard looks like below:
image


Chengxuan:
added batch size in the dashboard:
image

peterbroadhurst (approved):
This looks great @Chengxuan 🙌

PR #122, opened 2024-06-06
fix: deleting listeners from event stream pagination

EnriqueL8:
When deleting an event stream that has more listeners that the default pagination limit, it was failing due to incorrect logic of deleting the listener before using it as "offset" to query the next set of listeners to delete.

This PR fixes this by querying the next set before deleting the last one.

### DIFF
--- a/pkg/fftm/stream_management.go
+++ b/pkg/fftm/stream_management.go
@@ -32,7 +32,7 @@ import (
 	"github.com/hyperledger/firefly-transaction-manager/pkg/txhandler"
 )
 
-var (
+const (
 	startupPaginationLimit = 25
 )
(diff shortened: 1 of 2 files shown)


EnriqueL8:
Tested locally with a custom build of evmconnect in FF CLI

peterbroadhurst (changes requested):
We shouldn't need to duplicate code to fix the bug here

EnriqueL8:
Thanks for the pointer @peterbroadhurst - radically simpler now

peterbroadhurst (approved):
Thanks for the extra go-round to simplify @EnriqueL8

PR #123, opened 2024-06-09
FFTM new listener type for block notifications and receipt decoding functions

peterbroadhurst:
The FFTM basis for building up to the goals in FIR-18.
The FireFly core spelling that would expose this would be the next step on FIR-18, but I believe this PR can merge in isolation as the (useful in its own right) foundations in FFTM.

> There is one small FFCAPI structure changes that affects connectors/policy engines, but trivial to adopt.

The core proposal is to provide a reliable streaming interface for historical and current blocks, with the required number of confirmations, building on all the previous complex code for block-listening already built in EVMConnect etc.

There is definite code overlap with https://github.com/hyperledger/firefly-evmconnect/blob/8b0c3b46915aa8aab001a34bc03030d5ad44a381/internal/ethereum/blocklistener.go#L92 in handling confirmations, but after much back and forth I believe my proposal allows the two bits of work to operated in concert.

- EVMConnect - for EVM does the work of getting a stable stream of blocks at the head of the chain to shoulder-tap with hashes in an efficient order when there's a re-org
- This code - framework code that provides a historical playback of blocks, as well as interfacing with the head-listener, and layering on a confirmation threshold where anything before that is considered "final"

### Why blocks?

Because this is the only thing that chains can definitively give you as a stream efficiently to base a programming model around receipts with. The receipts are a side-effect of block confirmation, so if you want the receipts "in order" it's actually the blocks you need to listen for.

### Progress/plan in this PR

- [x] Work out architecture for what lets you listen for confirmed receipts in order (which is blocks :) )
- [x] Create core unit-tested logic for listening to blocks
- [x] Integrate the new listener type into event streams
- [x] Add FFCAPI for querying transaction receipt information, including ABI decoding `events` (aka `logs`) and `errors`
- [x] Test end to end with EVMConnect (no use of FF Core at this stage in the FIR-18 engineering)

## Also see:
- FIR: https://github.com/hyperledger/firefly-fir/pull/18
- EVMConnect: https://github.com/hyperledger/firefly-evmconnect/pull/141


pkg/ffcapi/block_info_by_number.go
```diff
@@ -22,6 +22,7 @@ import (
 
 type BlockInfoByNumberRequest struct {
 	BlockNumber        *fftypes.FFBigInt `json:"blockNumber"`
+	AllowCache         bool              `json:"allowCache"`
```
peterbroadhurst, line 25: Requires trivial change to `firefly-evmconnect` to propagate this to an existing internal boolean

peterbroadhurst:
Progress update on this:

Unit tested code is complete for integrating a new `type: "blocks"` (vs `type: "events") listener into the _existing_ event streams. So you can receive these in batches alongside your events and just ack them in the same way.

pkg/ffcapi/transaction_receipt.go
```diff
@@ -21,15 +21,25 @@ import (
...
-	TransactionHash string `json:"transactionHash"`
+	TransactionHash string            `json:"transactionHash"`
+	IncludeLogs     bool              `json:"includeLogs"`
+	EventFilters    []fftypes.JSONAny `json:"eventFilters"`
+	Methods         []fftypes.JSONAny `json:"methods"`
+	ExtractSigner   bool              `json:"extractSigner"`
 }
 
-type TransactionReceiptResponse struct {
+type TransactionReceiptResponseBase struct {
```
peterbroadhurst, line 31: As I worked this through in combination with https://github.com/hyperledger/firefly-evmconnect/pull/141, I found I had to make a tweak to the Receipt structure on the FFCAPI interface.

We have a (now slightly ugly looking sadly) JSON merging structure on the `EventWithContext` object, that defines how serialization happens into the JSON merging static and custom bits of the object together.

However, this is on `apitype` _not_ `ffcapi` where we just have `Event` - which is one part of what's merged.

Without doing really complex refactoring, I found a compromise where I map in the FFTM layer between the types. The easiest way to do this was adding this here.

internal/events/webhooks.go
```diff
@@ -103,12 +104,10 @@ func (w *webhookAction) attemptBatch(ctx context.Context, batchNumber int64, att
 	if w.isAddressBlocked(addr) {
 		return i18n.NewError(ctx, tmmsgs.MsgBlockWebhookAddress, addr, u.Hostname())
 	}
-	var resBody []byte
 	req := w.client.R().
 		SetContext(ctx).
 		SetBody(events).
-		SetResult(&resBody).
-		SetError(&resBody)
+		SetDoNotParseResponse(true)
```
peterbroadhurst, line 110: Tweaked after seeing this error with a webhook test framework:

```
FF21042: Webhook request failed: json: cannot unmarshal object into Go value of type []uint8 
```

internal/confirmations/confirmed_block_listener.go
```diff
@@ -0,0 +1,317 @@
...
+		if cbl.rollingCheckpoint != nil && cbl.rollingCheckpoint.Block >= cbl.fromBlock {
+			blockNumberToFetch = cbl.rollingCheckpoint.Block + 1
+		}
+		if len(cbl.blocksSinceCheckpoint) > 0 {
+			blockNumberToFetch = cbl.blocksSinceCheckpoint[len(cbl.blocksSinceCheckpoint)-1].BlockNumber.Uint64() + 1
+		}
+		cbl.stateLock.Unlock()
+
+		// Get the next block
+		nextBlock, err = cbl.bcm.getBlockByNumber(blockNumberToFetch, false, "")
```
Chengxuan, line 258: @peterbroadhurst I feel this could be too aggressive to avoid different block info with the same Block Number. The number of requests TM sends to a JSON-RPC endpoint affects performance a lot so we should keep it efficient.

The logic need to balance between the following choices when there are multiple block listeners in flight:
1. listeners should go and revalidate the information of each block from the "fromBlock" even if they are validating the same block at the same time.
2. listeners should use cached information of blocks when possible.

I can see the second one won't work with more logic added as we'll stuck at unmatching block parent hash when the cached block is outdated.

However, I do also feel the first one is a bit too aggressive if each listener will need to fetch each block with no cache at all. For example, if we have 10 listener, then we are emitting 10 get blockInfo call to the JSON-RPC endpoint for every single block.

Chengxuan: One solution could be default to use cache, and stop using cache when we detected the parent hash do not match the one in the unconfirmed previous block.

Chengxuan (commented):
LGTM, one comment around caching for getting next block

internal/confirmations/confirmed_block_listener.go
peterbroadhurst: I agree with all the desires and concerns of things to think about here @Chengxuan - so thank you for raising. 

I completely agree it would be great _if_ we could work out a way to have a cache on this line, but I actually I found I had to _disable the previously existing cache_ here because there's no way I could pre-empt knowing what re-org has happened on the node.

However, please note this **only kicks in** if the block addition by hash (which is cached) doesn't work. If we're just adding hashed block to the list as they arrive, none of the listeners will query at all. That's happening on the other go routine.

The optimization for block arriving in order from the chain to multiple listeners happens here (and is cache-enabled, because it's a get-by-hash which is deterministic to cache):
https://github.com/hyperledger/firefly-transaction-manager/blob/072aa386c57c784f02850948dd10dc0c6c04ff00/internal/confirmations/confirmed_block_listener.go#L196-L198

peterbroadhurst: Oh, just a minute. That comment of `// This block fits, and add this on the end.` reflects what I described above, but the actual code in the notification is only dealing with the re-org case.

Sorry - I must have forgotten to come back to that code to put in the add-to-end case in after I did some work.
I'll complete that, and then the code will work as I describe in my comment here.

You have also made me think of another optimization - the dispatcher (or "catch-up" routine) is always asking for blocks until it gets an error. This means in the caught-up notification cases, it's always doing an inefficient extra JSON/RPC call just to find out there's no new blocks.

If I maintain a "last notification height" on the notification listener (or similar), then we can optimize such that if the dispatch is caught up with the head of the chain, then there's no immediate re-poll necessary.

I'll do both of these and request a re-review.

peterbroadhurst: Ok - that's was harder than I'd thought! Logic got significantly more complex, to deal with the locking and avoid a situation where any go channel gets unnecessarily blocked.

However, I believe I now have it so that in the case the listener is keeping up, there will only be cached `eth_getBlockByHash` calls, and no `eth_getBlockByNumber` calls.

Chengxuan (approved):
Looks great. Thanks @peterbroadhurst

PR #125, opened 2024-07-09
Introduce a mode for transaction manager to be initiated as a golang module 

Chengxuan:
This PR separates the core functionality from the public interface choice (HTTP) which allows the core functionality provided by the transaction manager to be used as a golang module that backs up any format of the public interface or be remembered into another Golang project.

Here is a list of changes:
1. Introduced an extra parameter for creating a new transaction manager. When the extra parameter (module functions) is provided,  the created transaction manager will not initiate any HTTP server-specific components and allow events to be streamed back through the callback module function provided.
2. Exposed subsets of Stream, Listener and Transaction Handler functions as part of the module mode public interface.
3. the default transaction notifications are not exposed in module mode, use listeners instead


Chengxuan:
need to think through the set of interfaces

PR #134, opened 2025-02-19
add monitoring routes

Chengxuan:
As elaborated in https://github.com/hyperledger/firefly-common/pull/167, there is a need to expose more endpoints for operational checks. 

This PR propose a new `monitoring` sections in the configuration file for configurations of the monitoring server:
- whether to enable the monitoring server
- a `metricsPath`

The new monitoring server itself is a replacement for the existing metrics server.
- Expose the readiness and liveness check routes
- Expose the metrics routes

As a result the existing `metrics` section is marked as deprecated.


EnriqueL8 (commented):
Just need some clarification on a few things here:
- It looks like we don't use any of the work to build the server in FF Common in that PR referenced as we are using `httpserver` pkg and not `flap` pkg. We use the config spelling I believe.
- For some reason we have bundled a Golang version upgrade as part of this PR?
- Loads of linting changes as well

pkg/fftm/routes.go
```diff
@@ -28,11 +28,11 @@ func (m *manager) routes() []*ffapi.Route {
 		getEventStreamListener(m),
 		getEventStreamListeners(m),
 		getEventStreams(m),
-		getLiveStatus(m),
 		getStatus(m),
 		getSubscription(m),
 		getSubscriptions(m),
-		getReadyStatus(m),
+		getLiveStatus(m),  // TODO: remove this route from the API routes, they are already in the monitoring routes
+		getReadyStatus(m), // TODO: remove this route from the API routes, they are already in the monitoring routes
```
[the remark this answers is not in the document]
Chengxuan: yes, we should remove them as part of a major version update and it will introduce breaking changes.

Doesn't feel worth a major version update just for the sake of removing these 2 routes right now.

.github/workflows/go.yml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: golang:1.22-bullseye
+    container: golang:1.23-bullseye
```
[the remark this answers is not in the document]
Chengxuan: I don't believe FFTM is consumed by any repos other than `firefly-evmconnect`, which I can update as part of the PR to pick this up.

I'm not sure what needs to be coordinated between the other repos you listed above. Can you clarify why the coordination is needed?

pkg/fftm/api.go
```diff
@@ -59,24 +59,37 @@ func (m *manager) router(metricsEnabled bool) *mux.Router {
...
 
-	mux.NotFoundHandler = hf.APIWrapper(func(res http.ResponseWriter, req *http.Request) (status int, err error) {
+	mux.NotFoundHandler = hf.APIWrapper(func(_ http.ResponseWriter, req *http.Request) (status int, err error) {
 		return 404, i18n.NewError(req.Context(), i18n.Msg404NotFound)
 	})
 	return mux
 }
 
-func (m *manager) createMetricsMuxRouter() *mux.Router {
+func (m *manager) createMonitoringMuxRouter() *mux.Router {
```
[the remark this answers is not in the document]
Chengxuan: @EnriqueL8 The reason I didn't do it was because that felt like a larger piece of work. So I picked the approach of delaying that piece of work based on my current availability.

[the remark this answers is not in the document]
Chengxuan: I also do feel it will be easier for the ff-common owner to show an example of migrating a service (e.g. firefly core) to use the new common util, so we don't keep building up the debts.

.github/workflows/go.yml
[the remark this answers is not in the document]
EnriqueL8: In the past we have always coordinated version upgrades of all packages for a release and we do it from the bottom up. So we start at FireFly Common, Signer, etc...

Chengxuan, line 13:
```suggestion
    container: golang:1.22-bullseye
```

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hyperledger/firefly-transaction-manager
 
-go 1.22
+go 1.23
```
Chengxuan, line 3:
```suggestion
go 1.22
```

Makefile
```diff
@@ -17,7 +17,7 @@ coverage.html:
 		$(VGO) tool cover -html=coverage.txt
 coverage: test coverage.html
 lint: 
-		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
+		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.5
```
Chengxuan, line 20:
```suggestion
		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
```

pkg/fftm/api.go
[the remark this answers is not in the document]
EnriqueL8: So after digging the real reason we want to keep the initialization of the server separate is here https://github.com/hyperledger/firefly-transaction-manager/blob/b66a96a45b3b66d17d9aa913e5d2584c6da1da42/pkg/fftm/manager.go#L132-L134 

Somewhere in the TX handler it probably need the APIServer and the metrics get added

.github/workflows/go.yml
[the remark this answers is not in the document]
Chengxuan: yes. It doesn't feel like a technical thing, it's more about how the Firefly project treats dependency updates. I remember making the same comment to one of our contributors. But then I realized the current process doesn't encourage frequent version updates and probably because of that, most OSS projects would welcome non-major version upgrades in their repos to be done independently.

pkg/fftm/manager.go
```diff
@@ -90,13 +91,14 @@ func NewManager(ctx context.Context, connector ffcapi.API) (Manager, error) {
...
-		metricsServerDone: make(chan error),
-		metricsEnabled:    config.GetBool(tmconfig.MetricsEnabled),
-		eventStreams:      make(map[fftypes.UUID]events.Stream),
-		streamsByName:     make(map[string]*fftypes.UUID),
-		metricsManager:    metrics.NewMetricsManager(ctx),
+		connector:                connector,
+		apiServerDone:            make(chan error),
+		monitoringServerDone:     make(chan error),
+		deprecatedMetricsEnabled: config.GetBool(tmconfig.DeprecatedMetricsEnabled),
+		monitoringEnabled:        config.GetBool(tmconfig.MonitoringEnabled),
```
EnriqueL8, line 98: just a thought: why do we need to store both booleans? Can we just store that it's enabled and store the config section that we use is either the deprecated one or the new one. Then the code doesn't need all `or` checks

Chengxuan: when creat an HTTP server, if deprecatedMetricsEnabled, then use configuration under `metrics` section to initialised the HTTP server, if monitoringEnabled, then use the configuration under `monitoring` section to initialised the HTTP server.

Chengxuan: and when adding the metrics path (configurable) if deprecatedMetricsEnabled, then use the value of `metrics.path`, if monitoringEnabled, then use the value of `monitoring.metricsPath`.

EnriqueL8: Okay so after doing some more digging I see where the complexity occurs because of the interface with the APIServer. To create an API Server, you need to pass in a `config.Section` directly and it's easier in the case to just check is it one or the other and then pass that through. I was thinking we could store the `config.Section` as a variable in the manager and then just use that and we don't need all the switching inside the manager and just at the top when reading the config.

Ideally in ff-common, we should have a constructor which can accept a structure as config so we can handle breaking changes by reading in all the config in each component and then converting to that structure. We have done this in the past

EnriqueL8 (approved):
Thanks for the discussion on this one @Chengxuan

internal/apiclient/eventstreams.go
```diff
@@ -31,7 +31,7 @@ func (c *fftmClient) GetEventStreams(ctx context.Context) ([]apitypes.EventStrea
 		SetResult(&eventStreams).
 		Get("eventstreams")
 	if !resp.IsSuccess() {
-		return nil, fmt.Errorf(string(resp.Body()))
+		return nil, fmt.Errorf("%s", string(resp.Body()))
```
peterbroadhurst, line 34: This doesn't make sense to me.
Did you mean to address the lint issue that comes up with `fmt.Errorf` with a static string? If so, you should move to the `errors.New(text string)` function

EnriqueL8: I think this was done automatically by the golangci-lint?

peterbroadhurst: It should be raised tehre - but the recommendation is to use `errors.New()` I believe. I've never used some kind of "auto-fix" if that's what you're talking about, so I don't know what that would do.

My feedback remains

internal/persistence/postgres/transaction_writer.go
```diff
@@ -354,6 +355,7 @@ func (tw *transactionWriter) assignNonces(ctx context.Context, txInsertsByFrom m
 				}
 			}
 			log.L(ctx).Infof("Assigned nonce %s / %d to %s", signer, cacheEntry.nextNonce, op.txInsert.ID)
+			//nolint:gosec // Safe conversion as nextNonce is always positive
```
peterbroadhurst, line 358: Nit: This comment is inaccurate.

I agree with the ignore, but the justification is wrong (`uint64`-> `int64` cannot be "safe" due to positive - it can only be fully safe if you know all codepaths are < 2^63-1)

pkg/fftm/api.go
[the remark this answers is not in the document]
peterbroadhurst: Cannot review this until this is resolved. Sounds like some work needs to be done, but unclear what or by who.

@Chengxuan @EnriqueL8 - please let me know if you need me to action something.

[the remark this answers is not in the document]
Chengxuan: I'm thinking doing that one as a breaking change rather than introduce `/api` as an option for firefly common https://github.com/hyperledger/firefly-common/blob/main/pkg/ffapi/apiserver.go#L273

internal/apiclient/eventstreams.go
Chengxuan: I'll switch to errors.New(). It was fixed using code pilot.

pkg/fftm/api.go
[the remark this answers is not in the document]
Chengxuan: @peterbroadhurst i don't think you fully understand @EnriqueL8 's comment. He's comment was he thought that's the reason FFTM couldn't use the [NewAPIServer](https://github.com/hyperledger/firefly-common/blob/main/pkg/ffapi/apiserver.go#L99) function in FFTM. 

but it's not, and the comment was helpful if we were going to do the refactor.

The main reason the refactor is too big to fit in in because FFTM is now using `api` prefix whereas firefly common uses `api/v1`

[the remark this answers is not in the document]
Chengxuan: going to resolve this thread. 

Feel free to reopen with your request if you think this is unresolved

PR #139, opened 2025-06-22
Provide implementations access to TransactionHandler

peterbroadhurst:
The REST API of FFTM exposes straight forward `POST` APIs on top of the `TransactionHandler`:
https://github.com/hyperledger/firefly-transaction-manager/blob/7a882ddaeaf2e7b9b2203a053402576e436debd9/pkg/fftm/route__root_command.go#L86

However, there is no way for an implementation of FFTM (like EVMConnect) to get access to this interface to submit transactions directly, or use the other functions that they implement.

This PR adds a simple accessor to the existing interface.

It also adds `GetTransactionByIDWithStatus()` to enable idempotent submission.

### DIFF
--- a/pkg/fftm/manager.go
+++ b/pkg/fftm/manager.go
@@ -43,6 +43,7 @@ import (
 type Manager interface {
 	Start() error
 	Close()
+	TransactionHandler() txhandler.TransactionHandler
 }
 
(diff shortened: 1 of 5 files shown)


EnriqueL8 (approved):
Looks good

PR #141, opened 2025-07-03
Update to 1.5.5 FireFly Common and go 1.23.0 fixes

EnriqueL8:
Update to latest FireFly Common version and some go 1.23.0 fixes


internal/apiclient/fftm_client.go
```diff
@@ -19,7 +19,7 @@ package apiclient
 import (
 	"context"
 
-	"github.com/go-resty/resty/v2"
+	resty "github.com/go-resty/resty/v2"
```
matthew1001, line 22: Was there a reason for needing to qualify here, and in the other files e.g. `pkg/apitypes/ulid.go`?

matthew1001 (approved):
LGTM - one minor question

internal/apiclient/fftm_client.go
EnriqueL8: So with Go 1.23 it was complaining that it couldn't find that package, so I think there was a change to the import name. Normally it's the name of the package but it might be taking v2 now

PR #142, opened 2025-07-08
(WIP) Adding confirmation streaming in Event Stream

Chengxuan:
Adding support to stream confirmation events (the confirmation notification with a block event) for block listener only.

Unlike the built-in confirmation manager that's already used by the existing transaction handler, the confirmation events returned by **the event stream ALWAYS return a full set of confirmations** so that the consumer doesn't need to rebuild the full picture by accumulating them. This increases the size of the confirmation event, but hugely reduce the complexity of the client logic.

Breakdown of the details TBD


Chengxuan:
replaced by #143 which doesn't have breaking changes on existing API types

PR #145, opened 2025-08-06
Introduce update transaction API (PATCH /transactions/{transactionId})

Chengxuan:
The existing FFTM lacks the functionality of updating the values of transaction parameters. The values are completely controlled by the transaction handler to avoid unexpected outcomes due to user intervention.

However, for users who are familiar with the transaction lifecycle and have more advanced tuning requirements, an API allowing them to update the transaction parameter values directly is valuable, especially when TM is stuck in a situation that requires manual resolution. 

This PR introduces:
1. A Patch API for transactions,
2. A new function `HandleTransactionUpdate` in the Transaction Handler interface that needs to be implemented.
3. An implementation in the Simple Transaction Handler that handles transaction update **synchronously**


pkg/apitypes/managed_tx.go
```diff
@@ -179,6 +181,85 @@ func (mtx *ManagedTX) SetSequence(i int64) {
 	mtx.SequenceID = fmt.Sprintf("%.12d", i)
 }
 
+// ApplyExternalTxUpdates applies the external updates to the managed transaction
+// and returns the updates that were applied
+// ApplyExternalTxUpdates applies the external updates to the managed transaction
+// and returns the updates that were applied, whether any update occurred, and a map
+// of changed fields with their old and new values for easy JSON marshalling.
+func (mtx *ManagedTX) ApplyExternalTxUpdates(txUpdate TXUpdatesExternal) (txUpdates TXUpdates, updated bool, valueChangeMap map[string]map[string]string) {
```
peterbroadhurst, line 189: So the design I believe is that this is a helper function that policy implementations (which hold the complex threading model) can call back into, at the point the "task" of editing a transaction has made it to the right locked in-memory code that is safe to make the change.

Will validate this understanding through the rest of the review.

pkg/apitypes/managed_tx.go
```diff
@@ -179,6 +181,85 @@ func (mtx *ManagedTX) SetSequence(i int64) {
...
+	txUpdates = TXUpdates{}
+	valueChangeMap = make(map[string]map[string]string)
+
+	if txUpdate.To != nil && mtx.To != *txUpdate.To {
+		valueChangeMap["to"] = map[string]string{
```
peterbroadhurst, line 194: The `old`/`new` item needs to be an API structured type, rather than string fields please.

I haven't reached the point in understanding what this is used for:

> ... and a map
> // of changed fields with their old and new values for easy JSON marshalling.

pkg/apitypes/managed_tx.go
```diff
@@ -208,6 +289,15 @@ type TXUpdates struct {
 	ErrorMessage    *string           `json:"errorMessage,omitempty"`
 }
 
+type TXUpdatesExternal struct {
+	To              *string           `json:"to,omitempty"`
+	Nonce           *fftypes.FFBigInt `json:"nonce,omitempty"`
+	Gas             *fftypes.FFBigInt `json:"gas,omitempty"`
+	Value           *fftypes.FFBigInt `json:"value,omitempty"`
+	GasPrice        *fftypes.JSONAny  `json:"gasPrice,omitempty"`
+	TransactionData *string           `json:"transactionData,omitempty"`
```
peterbroadhurst, line 298: Please can we ensure we have a very clear knowledge (documented would be great in field-level docs) for how you clear this field.

Noting that means distinguishing between:
- Not changing the `transactionData`
- Setting the `transactionData` to be empty (simple transfer transaction)

We *must* ensure this works - even with the reference OSS handler.

The most common edit for experienced users fixing transaction that is broken, will be to turn it into a null transaction:
- Leaving the `nonce` unchanged
- Bumping the gas price artificially to allow a replacement transaction to go in
- Changing the `value` to be zero (if it wasn't before)
- Clearing the `transactionData`

This type of inert transaction that can be mined at very minimal execution transaction, is the safest way to clear bubbles where a transaction is breaking the pipe.

Noting a really sticky derivation of this made possible by this PR is:
1. Add a new transaction that is inert, but gets shoved at the **end of the nonce list**
2. Edit that transaction to have a `nonce` that fills a gap

pkg/fftm/transaction_management.go
```diff
@@ -85,6 +85,20 @@ func (m *manager) getTransactions(ctx context.Context, afterStr, limitStr, signe
 
 }
 
+func (m *manager) updateTransaction(ctx context.Context, txID string, txUpdate *apitypes.TXUpdatesExternal) (transaction *apitypes.ManagedTX, err error) {
+	if txUpdate == nil {
+		return nil, i18n.NewError(ctx, tmmsgs.MsgInvalidSortDirection)
+	}
+	updatedTx, err := m.txHandler.HandleTransactionUpdate(ctx, txID, *txUpdate)
```
peterbroadhurst, line 92: Ok - so the whole of the job of thread management etc. is delegated down into the policy engine implementation 👍

pkg/txhandler/simple/policyloop.go
```diff
@@ -272,13 +273,27 @@ func (sth *simpleTransactionHandler) pendingToRunContext(baseCtx context.Context
 	confirmNotify := pending.confirmNotify
 	receiptNotify := pending.receiptNotify
 	if syncRequest != nil {
-		ctx.SyncAction = *syncRequest
+		ctx.SyncAction = syncRequest.requestType
+		if syncRequest.requestType == ActionUpdate {
+			if syncRequest.txUpdates.GasPrice != nil {
+				return nil, i18n.NewError(ctx, tmmsgs.MsgTxHandlerUnsupportedFieldForUpdate, "gasPrice")
+			}
```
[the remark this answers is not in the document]
peterbroadhurst: Noting (per above discussions) this is a reference implementation, and the expectation is that providers of enterprise versions of FFTM are likely extending this simple engine with a much more sophisticated engine

peterbroadhurst (changes requested):
Some commentary for discussion on this great looking change before I'm confident it's ready to go in, plus just one explicitly request for a change where we had `"new"`/`"old"` strings being used rather than a Go structure... but more than that I didn't understand the purpose of that go structure.

EnriqueL8 (commented):
Thanks for this @Chengxuan - I would love to see a document explaining when a TX can be patched, what does it mean to resume/pause/suspend a transaction, overall what does the transaction lifecycle look like. We should put this in the main docs possibly under a new section.

peterbroadhurst (approved):
Thanks @Chengxuan - appreciate all the clarifications.

PR #150, opened 2025-11-07
return transaction receipt as part of the confirmation response

Chengxuan:
Adding the transaction receipt as part of the confirmation result so that the client does not need an extra step to retrieve the transaction receipt


peterbroadhurst (approved):
👍

PR #153, opened 2025-12-10
fix: duplicate notifications when walkChainForItem fails

EnriqueL8:
The was a bug in the way the notifications were processed where we end up with notifications not being removed.

When a new notification of type `EventLog`  is processed that is not at the head of the chain, we walk up the chain to verify it. We do this for each notification we have in memory to process in the confirmations manager loop. The existing code exits the loop (continue) if any of the process notifications fail. So let's say we fail to fetch a block for networking issues/rate limiting when walking up the chain for the Nth of M notifications, we don't mark the previous N notifications as successful but keep them in memory. 

This means the notifications are constantly being added and removed from the pending map, but not from the notification buffer that controls if things are inflight and can be dispatched to consumers. The problem gets worse when the getBlockByNumber fails more regularly and the buffer gets so large that eventually if it fails once in many tries is just stall the whole stream.

This PR attempts to fix this by returning which notifications were *not* processed, so the caller can clear those and just have the one in memory that do need processing reducing this problem. The clients in the connectors need to enable retry on the getBlockNumber calls to harden this.

There is also a potential issue if the block we are fetching to walk the chain is so old that is no longer available in the node wondering if @Chengxuan has any thoughts here if we should have a mode where if we are way behind and we don't have node that has the block we can optimistically confirm the event.

### DIFF
--- a/internal/confirmations/confirmations.go
+++ b/internal/confirmations/confirmations.go
@@ -423,12 +423,13 @@ func (bcm *blockConfirmationManager) confirmationsListener() {
 
 		// Process any new notifications - we do this at the end, so it can benefit
 		// from knowing the latest highestBlockSeen
-		if err := bcm.processNotifications(notifications, blocks); err != nil {
-			log.L(bcm.ctx).Errorf("Failed processing notifications: %s", err)
+		// This returns the notifications that were not processed successfully so we store them
+		var err error
+		notifications, err = bcm.processNotifications(notifications, blocks)
+		if err != nil {
+			log.L(bcm.ctx).Errorf("Failed processing notifications: %d due to error %s", len(notifications), err)
 			continue
 		}
-		// Clear the notifications array now we've processed them (we keep the slice memory)
-		notifications = notifications[:0]
 		scheduleAllTxReceipts := !receivedFirstBlock && blockHashCount > 0
 		// Mark receipts stale after duration
 		bcm.scheduleReceiptChecks(scheduleAllTxReceipts)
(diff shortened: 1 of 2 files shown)


internal/confirmations/confirmations.go
```diff
@@ -423,12 +423,13 @@ func (bcm *blockConfirmationManager) confirmationsListener() {
 
 		// Process any new notifications - we do this at the end, so it can benefit
 		// from knowing the latest highestBlockSeen
-		if err := bcm.processNotifications(notifications, blocks); err != nil {
-			log.L(bcm.ctx).Errorf("Failed processing notifications: %s", err)
+		// This returns the notifications that were not processed successfully so we store them
+		var err error
+		notifications, err = bcm.processNotifications(notifications, blocks)
+		if err != nil {
+			log.L(bcm.ctx).Errorf("Failed processing notifications: %d due to error %s", len(notifications), err)
```
Chengxuan: Can we extend the log here to include information similar to what's in the `Confirmation listener processed %d block hashes` logs below, so we can still get a summary of the process progress?

EnriqueL8: 100% - let me do that

Chengxuan:
> There is also a potential issue if the block we are fetching to walk the chain is so old that is no longer available in the node

This would be a problem if user configures an listener with a fromBlock that no longer in the targeting RPC endpoint. OR user restarting transaction manager after a long time and the checkpoint block is no longer available.

However, the event listener will detect the problem first and then get stuck in a broken state on query block by range. 

So for now, we give users the choice to decide what to do with those listeners. And for these situations, it feels important for user to know the situation, "re-using" the same listener knowing it will lose event in between feels dangerous.

Chengxuan (approved):
@EnriqueL8 if I understand correctly, this PR solves a problem that: already confirmed pending items now get removed properly, so they don't get replayed on each loop iteration. But it doesn't solve the problem of `walkChainForItem` is more prone to json-rpc endpoint errors due to lack of retry by default.

It worth considering flipping the default value of https://github.com/kaleido-io/firefly-evmconnect/blob/main/config.md#connectorretry on 429s at least.

EnriqueL8:
@Chengxuan agreed, but that is up to each connector and not in this toolkit

PR #155, opened 2025-12-11
fix: do not recreate a new notifications slice

EnriqueL8:
We shouldn't be changing the notification array memory location as it's being read in a select statement 

`case notification := <-bcm.bcmNotifications:`

### DIFF
--- a/internal/confirmations/confirmations.go
+++ b/internal/confirmations/confirmations.go
@@ -490,7 +490,7 @@ func (bcm *blockConfirmationManager) processNotifications(notifications []*Notif
 	}
 
 	// Clear the notifications slice
-	return []*Notification{}, nil
+	return notifications[:0], nil
 }
 
 func (bcm *blockConfirmationManager) dispatchReceipt(pending *pendingItem, receipt *ffcapi.TransactionReceiptResponse, blocks *blockState) {


matthew1001:
It sounds like this addresses the issue, just to clarify it sounds in the PR description that returning a new slice affects the `select` statement in the next iteration of the `for` loop?

EnriqueL8:
Yeah so this function `processNotifications` is called in the loop but this loop has a select statement at the top when it gets notified of new notifications or new block hashes  https://github.com/hyperledger/firefly-transaction-manager/blob/ef44d517b899f98a89b88a96e4f5c69b62879ddc/internal/confirmations/confirmations.go#L389

I'm reverting to the existing behaviour we had here https://github.com/hyperledger/firefly-transaction-manager/blob/ef44d517b899f98a89b88a96e4f5c69b62879ddc/internal/confirmations/confirmations.go#L431

We moved to return the notifications that were not processed in this PR https://github.com/hyperledger/firefly-transaction-manager/pull/153

matthew1001 (approved):

Chengxuan (approved):

EnriqueL8:
Thanks both - going to keep a close eye on the test for this one

PR #156, opened 2026-01-02
adding transaction listing index

Chengxuan:
This PR adds 2 indices that helps on transaction listing when the number of transaction rows grows significantly:

- a composite index of from,status and nonce for the transaction handler to fetch unprocessed transactions for a give "from" address
- an index on `created` for time window based list queries

# NOTE: for a DB with a large volume of **PENDING** transaction records, this new migration could take a long time to run


EnriqueL8 (commented):
Thanks @Chengxuan I wonder if this could be an optional additive thing that a user could opt-in to have - instead of an index that consumes a lot of storage cost in PSQL

Chengxuan:
Adding index itself doesn't solve the performance issue. A refactor of the db table is required.

PR #159, opened 2026-04-21
introduce a lightweight tracking mode for connectors and update confirmation manager to cope with it

Chengxuan:
## Problem

for https://github.com/hyperledger/firefly-evmconnect/issues/191 and https://github.com/hyperledger/firefly-transaction-manager/issues/161

> When using EVM Connect, downloading detailed block information to maintain an in-memory partial chain to track the latest block progress can be very costly. Especially, the number of transactions in each block is high. Therefore, for users who don't need to index block information and only want to use transaction submission and event listener features, it's a big overhead

Therefore, a lightweight mode is proposed to maintain the usability of transaction submission and event listeners.

## Solution

This PR contains

- a major FireFly Connector API (FFCAPI) interface update to include a function for connectors to return the tracking mode that the connector is running in: `GetBlockListenerTrackingMode`
  - `inMemoryPartialChain`, this is the current tracking mode, the connector builds an in-memory partical chain to track the block progress of the blockchain. This mode downloads information of each block from the chain, causing more data transportation, to provide the benefit of a verbose record of block details. 
  - `headBlockNumber`, this is the new tracking mode added, which is a light weight mode compared to the existing `inMemoryPartialChain` mode, the connector only records the head block number to track block progress. **Note: this is an optional mode to opt in, the connector developer can decide whether it makes sense to add it** An example implementation is proposed in EVM connector: https://github.com/hyperledger/firefly-evmconnect/pull/190. EVM Connector also added a configuration ( `connector.blockTrackingMode` ) so that users can control which mode they'd like the connector to run under.
- Updated two components in the transaction manager to support the new `headBlockNumber` mode due to reduced information in that mode:
  - Confirmation manager: When the connector is in `headBlockNumber`, it will no longer be able to return a list of block information. Therefore, to surface the confirmation progress, a new field `currentConfirmationCount` has been introduced (alongside the existing `targetConfirmationCount` field) **NOTE: it's possible to introduce a separate configuration (e.g. confirmations.confirmationMode) for confirmation manager to omit detailed block information when the connector is running in `inMemoryPartialChain`, but I don't see the need for it for now** See the section below for details.
  - Block listener: as the sole purpose of the block listener is to index the block with details, it counter efficient to support it when the connector are not provide efficient caching for block details. It's disallowed when a connector is running under `headBlockNumber` tracking mode 

## Details about confirmation manager changes

Add a confirmation logic that only needs the chain head block number to avoid downloading full block headers / building an in-memory partial chain.

That keeps confirmation work cheap and reduces the amount of unnecessary data downloaded from the chain.

### Implementation

When the block listener is in the headBlockNumber mode, the confirmation check is derived from the head minus the receipt block number. When the required depth is reached, validity is checked by re-fetching the transaction receipt and comparing its block hash to the pending item ( this means re-org / fork handling leans on receipt reconciliation against head-driven confirmation counts rather than walking downloaded block metadata).

### Example confirmation output in two modes:

Existing mode (list of blocks built using in memory chain)

```
 {
  "confirmations": [
    {
      "blockHash": "0xxxxxxxxxx1",
      "blockNumber": "1",
      "parentHash": "0xxxxxxxxxx0"
    },
    {
      "blockHash": "0xxxxxxxxxx2",
      "blockNumber": "2",
      "parentHash": "0xxxxxxxxxx1"
    }
  ],
  "confirmed": true,
  "currentConfirmationCount": 1,
  "targetConfirmationCount": 1
}
```

Light weight mode (subset information):

```
 {
  "confirmed": true,
  "currentConfirmationCount": 1,
  "targetConfirmationCount": 1
}
```


pkg/ffcapi/api.go
```diff
@@ -104,7 +116,9 @@ type ConfirmationUpdateResult struct {
 	// in the in-memory partial chain
 	// WARNING: mutation to this list is not expected, invalid modifications will cause inefficiencies in the reconciliation process
 	//          `rebuilt` will be true if an invalid confirmation list is detected by the reconciliation process
-	Confirmations           []*MinimalBlockInfo         `json:"confirmations,omitempty"`
+	Confirmations            []*MinimalBlockInfo `json:"confirmations,omitempty"`  // the current list of confirmations for this reconcile request, only returned when confirmation mode is set to validatedBlocks
```
hosie: What is `confirmation mode` and `validatedBlocks`?  Was this previous spelling for `BlockListenerTrackingMode` and `inMemoryPartialChain`

hosie (commented):
LGTM with one comment.

pkg/ffcapi/api.go
Chengxuan: yes, thanks for catching it, will update the comment

internal/confirmations/confirmations.go
```diff
@@ -114,12 +117,16 @@ func NewBlockConfirmationManager(baseContext context.Context, connector ffcapi.A
...
-		requiredConfirmations: config.GetInt(tmconfig.ConfirmationsRequired),
-		staleReceiptTimeout:   config.GetDuration(tmconfig.ConfirmationsStaleReceiptTimeout),
-		bcmNotifications:      make(chan *Notification, config.GetInt(tmconfig.ConfirmationsNotificationQueueLength)),
-		pending:               make(map[string]*pendingItem),
-		newBlockHashes:        make(chan *ffcapi.BlockHashEvent, config.GetInt(tmconfig.ConfirmationsBlockQueueLength)),
-		metricsEmitter:        cme,
+		requiredConfirmations: config.GetUint64(tmconfig.ConfirmationsRequired),
+		// currently the confirmation result is driven by the block listener mode,
+		// because when the block listener is in the headBlockNumber mode, the block information is not fetched and cached by the connector
+		// Therefore, it will be super inefficient to build the in memory partial chain in confirmations manager
```
peterbroadhurst: Don't understand this comment sorry @Chengxuan 

By `currently` do you mean there's a TODO? Or was this written before you implemented your change and you need to update the comment?

internal/confirmations/confirmations.go
```diff
@@ -406,9 +414,12 @@ func (bcm *blockConfirmationManager) confirmationsListener() {
 		blocks := bcm.newBlockState()
 
 		if bcm.blockListenerStale {
-			if err := bcm.walkChain(blocks); err != nil {
-				log.L(bcm.ctx).Errorf("Failed to walk chain after restoring blockListener: %s", err)
-				continue
+			if bcm.blockListenerTrackingMode == ffcapi.BlockListenerTrackingModeInMemoryPartialChain {
```
peterbroadhurst: `BlockListenerTrackingModeInMemoryPartialChain` is very wordy

What about:
- Option name: `ConfirmationMode` 
- Option1: `FullBlockVerification` - the default, we check every block ourselves in memory and build the chain so it's presented as evidence back to the user
- Option2: `BlockNumber` - the new option, we just compare the numbers

pkg/ffcapi/api.go
```diff
@@ -95,6 +98,15 @@ type API interface {
...
 }
 
+type BlockListenerTrackingMode string
+
+const (
+	// BlockListenerTrackingModeHeadBlockNumber in this mode, the block listener tracks the head block number of the blockchain only without fetching any block details
+	BlockListenerTrackingModeHeadBlockNumber BlockListenerTrackingMode = "headBlockNumber"
+	// BlockListenerTrackingModeInMemoryPartialChain in this mode, the block listener tracks the in-memory partial chain of blocks, fetching block details as needed
+	BlockListenerTrackingModeInMemoryPartialChain BlockListenerTrackingMode = "inMemoryPartialChain"
+)
```
peterbroadhurst, line 108: My spelling comments/suggestions are above (probably would have been better here, sorry)

internal/confirmations/confirmations.go
peterbroadhurst: Other spelling options to provoke thinking:
- `VerificationLevel` - `block_hashes`, `block_number`
- `ConfirmationLevel` - `disabled`, `number_only`, `full_confirmation_chain`
- `ReceiptConfirmationChain` - `enabled`/`disabled`

peterbroadhurst: I also think the PR header should clearly express the externals differences between the receipts you'll get - I don't remember how this pops out, but I think it's in the FFTM layer rather than the connector how the receipt is shaped.

I assume we'll return `confirmationCount: 5` in the new mode, rather than `confirmations: [{...}]` in the full verification mode we have today.

Chengxuan: @peterbroadhurst I can see your confusion as the existing description spent lots of words on the confirmation manager changes. But the key feature introduced is actually a light weight chain blocks tracking mode.

I've updated the main description with more background and the reasoning. Can you take another look and share your thoughts

internal/confirmations/confirmations.go
Chengxuan: No, it's a decision. I'll clean up `currently`

internal/confirmations/confirmations.go
peterbroadhurst: The main description looks really nice.

At the end you naturally explain it as:

> - Existing mode (list of blocks built using in memory chain)
> - Light weight mode (subset information):

That leads me to believe a directly way to respond to my feedback would be:

- `confirmationMode: [light] | [full]`

It certainly seems nicer than my suggestions.

internal/confirmations/confirmations.go
```diff
@@ -406,9 +411,12 @@ func (bcm *blockConfirmationManager) confirmationsListener() {
 		blocks := bcm.newBlockState()
 
 		if bcm.blockListenerStale {
-			if err := bcm.walkChain(blocks); err != nil {
-				log.L(bcm.ctx).Errorf("Failed to walk chain after restoring blockListener: %s", err)
-				continue
+			if bcm.chainTrackingMode == ffcapi.ChainTrackingModeFull {
+				// only need to build the in memory partial chain when the block listener supports it
```
Chengxuan:
```suggestion
				// only need to build the in memory partial chain when the block details are available
```

internal/confirmations/confirmations.go
```diff
@@ -548,6 +556,12 @@ func (bcm *blockConfirmationManager) removeItem(pending *pendingItem, stale bool
 }
 
 func (bcm *blockConfirmationManager) processBlockHashes(blockHashes []string) {
+	if bcm.chainTrackingMode == ffcapi.ChainTrackingModeLight {
+		// for headBlockNumber mode, we don't need to fetch blocks to form a local in memory partial chain
+		// we just need to do confirmation check and dispatch confirmations for any applicable pending items
```
Chengxuan:
```suggestion
		// for light chain tracking mode, no block details are available, only need to calculate the number of confirmations using head block number
```

internal/confirmations/confirmations.go
```diff
@@ -89,14 +90,16 @@ type blockConfirmationManager struct {
...
 	connector             ffcapi.API
 	blockListenerStale    bool
 	metricsEmitter        metrics.ConfirmationMetricsEmitter
-	requiredConfirmations int
+	requiredConfirmations uint64
+	chainTrackingMode     ffcapi.ChainTrackingMode
 	staleReceiptTimeout   time.Duration
 	bcmNotifications      chan *Notification
 	highestBlockSeen      uint64
+	headBlockNumber       uint64 // used for confirmation when block listener is in headBlockNumber mode
```
Chengxuan:
```suggestion
	headBlockNumber       uint64 // used for confirmation when the connector is running in light mode
```

peterbroadhurst (approved):
Marking approval, as @hosie and @Chengxuan discussed this and reached a joint reasoned proposal on the spelling of:

- The setting is `connector.chainTrackingMode`, and is enabled on the the specific connector implementation (`firefly-evmconnect` in this case)
- The two possible values are:
    - `full`
    - `light`

These values are described in code comments on the FFTM side:

https://github.com/kaleido-io/firefly-transaction-manager/blob/852459b5a29d758290410fb27617a66d6fd28cf6/pkg/ffcapi/api.go#L104-L107

pkg/ffcapi/api.go
peterbroadhurst: This was resolved per the approval comments

internal/confirmations/confirmations.go
peterbroadhurst: This was resolved per the approval comments.

PR #160, opened 2026-04-21
[metrics] Expose MetricsRegistry for Custom Metrics Downstream

hosie:
So that we can enable ffresty client metrics, and add more metrics for clients of fftm.


Chengxuan (approved):

Chengxuan (approved):
LGTM , this was raised by @onelapahead in https://github.com/hyperledger/firefly-transaction-manager/pull/157 but got accidentally closed

PR #182, opened 2026-08-30
Update connector contract to protect delivery-ack window with certainty

peterbroadhurst:
This fixes a scenario where an event stream listener could silently drop events.
The scenario has been observed in practice when there are multiple listeners on a single event stream, and those listeners are catching up to the head while redelivery is occurring delaying acknowledgment of events detected during catchup by one of those listeners.

Each listener's checkpoint within the overall event stream advances two ways:

1. From an event delivered and acked by the application
2. From FFTM poling the collector with `EventListenerHWM` to check the position the connector's **scan** has reached

The second can be ahead of the first, as the connector pushes events into a channel in FFTM and continues to scan forwards to find more events while FFTM assembles those events into a batch, and delivers that batch awaiting ack.

The guards against moving the checkpoint forwards while those events were pending acknowledgment were not strong enough in FFTM.

## Fix architecture

A minimal **additive** change is made to the FFCAPI on the return value to `EventListenerHWMResponse()`, so that it not only returns the scan position, but also the checkpoint position of the highest event passed to the FFTM channel.

When a connector implements that simple contract (separate PR to EVMConnect), the changes in this FFTM PR provide an absolute assurance that the checkpoint is never moved forward by the scan interval, while there are uncommitted deliveries that have been pushed from the connector to FFTM.

https://github.com/hyperledger-firefly/transaction-manager/blob/5e887b9fb1037ceaae58af81c307b89a47240976/pkg/ffcapi/event_listener_hwm.go#L28-L58

Also included:

- **Block listeners** no longer take the HWM route. They're handled inside FFTM's own confirmation manager, so we were querying the connector with listener IDs it has never heard of.
- **Never silently drop** An event arriving behind an HWM-derived checkpoint is now delivered with a warning. We can't distinguish it from a genuine re-detection, and a duplicate is recoverable
    - Note FireFly core de-duplicates on protocol ID, so this does not mean application redelivery in practice

### DIFF
--- a/pkg/ffcapi/event_listener_hwm.go
+++ b/pkg/ffcapi/event_listener_hwm.go
@@ -25,7 +25,34 @@ type EventListenerHWMRequest struct {
 	ListenerID *fftypes.UUID `json:"listenerId"`
 }
 
+// EventListenerHWM return value.
+//
+// Important:
+//
+// The connector SHOULD implement LastDetected as well as Checkpoint on the response returned.
+//
+// Checkpoint is how far the connector has scanned.
+// LastDetected is the checkpoint of the last event it pushed to FFTM.
+//
+// These differ, because the connector can scan past events that are still sat in the
+// channel unread - FFTM cannot see those for itself, so it must be told.
+//
+// FFTM applies Checkpoint only once everything up to LastDetected has been delivered and acked.
+//
+// The connector needs to hold in memory a LastDetected value (to return when asked) that
+// is updated BEFORE updating the value it returns for the scan position.
+// When building the response for EventListenerHWM() read the scan position first, then
+// the last detected (or guard them both under a single mutex).
+//
+// Following these simple rules ensures FFTM never moves the checkpoint past an event
+// that is still in flight.
+//
+// Setting nil LastDetected is correct when nothing has been pushed for this listener.
+//
+// A connector that never implements LastDetected (always nil) is functional, but accepts
+// a risk of a checkpoint being persisted past events FFTM has not yet acknowledged.
 type EventListenerHWMResponse struct {
-	Checkpoint EventListenerCheckpoint `json:"checkpoint"`
-	Catchup    bool                    `json:"catchup,omitempty"` // informational only - informs an operator that the stream is catching up
+	Checkpoint   EventListenerCheckpoint `json:"checkpoint"`             // how far the connector has scanned
+	Catchup      bool                    `json:"catchup,omitempty"`      // informational only - informs an operator that the stream is catching up
+	LastDetected EventListenerCheckpoint `json:"lastDetected,omitempty"` // the checkpoint of the last event it pushed to FFTM
 }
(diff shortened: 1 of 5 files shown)


pkg/ffcapi/event_listener_hwm.go
```diff
@@ -25,7 +25,34 @@ type EventListenerHWMRequest struct {
...
+// Setting nil LastDetected is correct when nothing has been pushed for this listener.
+//
+// A connector that never implements LastDetected (always nil) is functional, but accepts
+// a risk of a checkpoint being persisted past events FFTM has not yet acknowledged.
 type EventListenerHWMResponse struct {
-	Checkpoint EventListenerCheckpoint `json:"checkpoint"`
-	Catchup    bool                    `json:"catchup,omitempty"` // informational only - informs an operator that the stream is catching up
+	Checkpoint   EventListenerCheckpoint `json:"checkpoint"`             // how far the connector has scanned
+	Catchup      bool                    `json:"catchup,omitempty"`      // informational only - informs an operator that the stream is catching up
+	LastDetected EventListenerCheckpoint `json:"lastDetected,omitempty"` // the checkpoint of the last event it pushed to FFTM
```
Chengxuan, line 60: Maybe `DetectedCheckpoint` / `PushedCheckpoint` could be a better name?

pkg/ffcapi/event_listener_hwm.go
```diff
@@ -25,7 +25,34 @@ type EventListenerHWMRequest struct {
...
 }
 
+// EventListenerHWM return value.
+//
+// Important:
+//
+// The connector SHOULD implement LastDetected as well as Checkpoint on the response returned.
+//
+// Checkpoint is how far the connector has scanned.
+// LastDetected is the checkpoint of the last event it pushed to FFTM.
```
Chengxuan, line 35: Worth highlighting `Checkpoint >= LastDetected` here.

// Checkpoint is how far the connector has scanned.
// LastDetected is how far the connector has pushed the scanned events to FFTM

Chengxuan (approved):
The incremental changes look great to me. A few suggestions on the doc and naming of the variable.

pkg/ffcapi/event_listener_hwm.go
peterbroadhurst: Wondering how strongly you feel about this one?

Of the three I personally like `lastDetected` most as it implies:

- It's not a checkpoint. It's the same data (block height etc.) that _could become_ a checkpoint
- It describes it's the fact that it's been detected, nothing more.

Chengxuan: My initial reaction is `LastDetected` is a timestamp. Then I noticed they all shared the same type `EventListenerCheckpoint` and it's not a primitive type. So that's where my preference for having `checkpoint` in the name comes from.

Chengxuan: But it's a minor point, reflected by my approval. After reading the code and comments, they made sense to me after I validated few doubts through code. The most important one is the other suggestion in the code comment. It helped me to validate that `LastDetected` is a historical value of `Checkpoint` and `Checkpoint` do not roll back.

PR #186, opened 2026-09-11
sort queued items before dispatching in light mode for non-zero confirmation

Chengxuan:
## Summary

In light chain tracking mode with `confirmationsRequired > 0`, events dispatched from the confirmation manager weren't guaranteed to be in block order. This could occasionally lead to a later-arriving event being treated by the event stream as a re-detection of an earlier one, so it wasn't forwarded. This path isn't used when `confirmationsRequired = 0` or during catch-up.

## Root cause

`checkAndDispatchConfirmationsUsingBlockHeight` builds its list of newly-confirmable items from `bcm.pending`, a Go map. Map iteration order isn't guaranteed, so when a head block update confirmed several pending events at once, they were dispatched in map order rather than block order.

Full chain tracking mode already handles this correctly in `processBlock` via `sort.Sort(notifications)` before dispatch. Light mode was missing the equivalent sort.

## Fix

Sort pending items by block number, transaction index, and log index (using the existing `pendingItems` sort type) before dispatching in `checkAndDispatchConfirmationsUsingBlockHeight`, bringing it in line with full mode.

## Testing

- Added `TestBlockConfirmationManagerHeadBlockNumberDispatchesInBlockOrder`, which confirms a batch of light-mode events together via one head block update and asserts dispatch order. It reproduced the ordering issue consistently before the fix and passes consistently after.
- `go build ./...`, `go vet ./...`, and the `internal/confirmations` suite all pass.
- The full `go test ./...` run has two unrelated pre-existing failures in `persistence/dbmigration` and `persistence/postgres` (a `golang-migrate` panic in this environment), present on `main` before this change too.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/confirmations/confirmations.go
+++ b/internal/confirmations/confirmations.go
@@ -846,7 +846,7 @@ func (bcm *blockConfirmationManager) walkChainForItem(pending *pendingItem, bloc
 	}
 
 	if bcm.chainTrackingMode == ffcapi.ChainTrackingModeLight {
-		return bcm.confirmationCheckUsingHeadBlockNumber(pending)
+		return bcm.dispatchConfirmationUsingHeadBlockNumber(pending)
 	}
 
 	pendingKey := pending.getKey()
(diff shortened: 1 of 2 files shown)


internal/confirmations/confirmations.go
```diff
@@ -895,21 +895,27 @@ func (bcm *blockConfirmationManager) walkChainForItem(pending *pendingItem, bloc
...
 		items = append(items, p)
 	}
 	headBlock := bcm.headBlockNumber
 	bcm.pendingMux.Unlock()
+	// bcm.pending is a map, so iterating it gives us no ordering guarantee. A single head block
+	// update can push many pending items over the confirmation threshold at once, so - as with
+	// processBlock's notifications in full chain tracking mode - we must sort by block order
+	// before dispatching, or events can be delivered out of order (and then dropped downstream
+	// as apparent re-detections once the checkpoint moves past them).
+	sort.Sort(items)
```
peterbroadhurst, line 909: Cool the ordering is here:
https://github.com/hyperledger-firefly/transaction-manager/blob/5915cbc4e0e30dea25068b770dadbcc8bdfa9321/internal/confirmations/confirmations.go#L213-L225

peterbroadhurst (approved):

